PR #251, opened 2019-02-15
Groups from ldap and required group for api calls

### LINKED ISSUE #250: Grab group memberships from ldap and add a group requirement for API
Use the `populate_user` signal from `django_auth_ldap.backend` to build group memberships for a users which is successfully authenticated.

Also add a permission class to require a specific group to do API calls. First edition will be to limit all API calls to a single group, but later add more granuality.

Inital work for #67.

oyvindkolbu:
Resolves #250.


mregsite/settings.py
```diff
@@ -132,13 +138,14 @@
...
 STATIC_URL = '/static/'
+STATIC_ROOT = os.path.join(BASE_DIR, 'static/')
 
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': (
         'mreg.authentication.ExpiringTokenAuthentication',
     ),
     'DEFAULT_PERMISSION_CLASSES': (
-        'rest_framework.permissions.IsAuthenticated',
+        'mreg.api.permissions.IsInRequiredGroup',
```
oyvindkolbu: Could possible set it in local_settings.py instead. Though with REQUIRED_USER_GROUP unset it will allow any authenticated user...

mreg/api/permissions.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Allows only access to users in the required group.
+    """
+
+    def has_permission(self, request, view):
+        if not bool(request.user and request.user.is_authenticated):
+            return False
+        REQUIRED_USER_GROUP = getattr(settings, 'REQUIRED_USER_GROUP', None)
+        if REQUIRED_USER_GROUP is None:
+            return True
```
oyvindkolbu: Sane to `return True` here? Maybe raise an error.

ponas: I'd raise an error

oyvindkolbu:
Will also need to fix testing...
```
Ran 140 tests in 2.663s
FAILED (failures=92, errors=9)
```

PR #266, opened 2019-03-01
Add API to logout a token. Actually just delete it.

### LINKED ISSUE #265: Add a logout method for tokens
Currently the tokens are valid until they expire:
https://github.com/usit-gd/mreg/blob/22e21e764103a109515ea37b28182e2094a7f029/mreg/authentication.py#L21-L22

Add a logout API to force an immediate invalidation of the token.

oyvindkolbu:
Resolves #265.

### DIFF
--- a/mreg/api/urls.py
+++ b/mreg/api/urls.py
@@ -3,6 +3,6 @@ from django.urls import path
 from . import views
 
 urlpatterns = [
-    path('logout/', views.Logout.as_view()),
+    path('token-logout/', views.TokenLogout.as_view()),
     path('token-auth/', views.ObtainExpiringAuthToken.as_view()),
 ]
--- a/mreg/api/urls.py
+++ b/mreg/api/urls.py
@@ -3,5 +3,6 @@ from django.urls import path
 from . import views
 
 urlpatterns = [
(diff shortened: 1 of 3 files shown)


mreg/api/urls.py
```diff
@@ -3,5 +3,6 @@
 from . import views
 
 urlpatterns = [
+    path('logout/', views.Logout.as_view()),
```
oyvindkolbu: "logout" or "token-logout"? -- futureproofing...

mreg/api/views.py
```diff
@@ -20,3 +21,10 @@ def post(self, request, **kwargs):
 
             return Response({'token': token.key})
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class Logout(APIView):
```
oyvindkolbu: Ditto. Logout -> TokenLogout?

oyvindkolbu:
See questions in code.

mreg/api/urls.py
ponas: token-logout!

mreg/api/views.py
ponas: TokenLogout!

PR #270, opened 2019-03-04
Dhcpv6 by v4 and updated at for hosts and ipaddresses

oyvindkolbu:


mreg/api/v1/views.py
```diff
@@ -279,6 +280,12 @@ def patch(self, request, *args, **kwargs):
             return Response(status=status.HTTP_204_NO_CONTENT, headers={'Location': location})
 
 
+@api_view(['GET'])
+def host_newest(request, *args, **kwargs):
+    newest = Host.objects.order_by('-updated_at').first()
+    return Response(HostSerializer(newest).data)
```
oyvindkolbu: Not sure about this one. Could have used [DRF pagination](https://www.django-rest-framework.org/api-guide/pagination/) to support a `page_size` / `limit` keyword, but it would break every API usage as the json-output would be changed to a new format:
```
GET https://api.example.org/hosts/?ordering=-updated_at&limit=1

HTTP 200 OK
{
    "count": 1234,
    "next": "https://api.example.org/hosts/?limit=1&ordering=-updated_by&page=2",",
    "previous": null,
    "results": [
       …
    ]
}
```

oyvindkolbu:
Splitting this pull into a dedicated for dhcpv6-by-v4, and pagination solution for the "newest" Host and Ipaddress objects.

PR #273, opened 2019-03-05
Add pagination to mreg

oyvindkolbu:
Will break mreg-cli.


ponas:
Should all list-like be paginated for consistency?

oyvindkolbu:
> Should all list-like endpoints be paginated for consistency?

Probably. More churn, but should just bite the bullet.

PR #278, opened 2019-03-25
Sshfp support

dmykarp:
Still reading on the topic if there could be some unsupported combination of algorithm and digests. Will add tests if there are.


dmykarp:
RFCs don't mention any restrictions on combinations. Any allowed value can be used with any other allowed value.

mreg/api/v1/zonefile.py
```diff
@@ -98,6 +98,18 @@ def mx_zf_string(self, name, ttl, priority, mx):
...
 
+    def sshfp_zf_string(self, name, ttl, algorithm, hash_type, fingerprint):
+
+        data = {
+            'name': name,
+            'ttl': ttl,
+            'record_type': "SSHFP",
+            'algorithm': algorithm,
+            'hash_type': hash_type,
+            'fingerprint': f'{fingerprint}'
```
oyvindkolbu: Uneeded f-string.

mreg/models.py
```diff
@@ -327,6 +327,20 @@ def loc_string(self, zone):
...
+    ttl = models.IntegerField(blank=True, null=True, validators=[validate_ttl])
+    algorithm = models.IntegerField(choices=((1, 'RSA'), (2, 'DSS'), (3, 'ECDSA'), (4, 'Ed25519')))
+    hash_type = models.IntegerField(choices=((1, 'SHA-1'), (2, 'SHA-256')))
+    fingerprint = models.CharField(max_length=64, validators=[validate_hexadecimal])
+
+    class Meta:
+        db_table = 'sshfp'
+
+    def __str__(self):
+        return str(self.host)
```
oyvindkolbu: Better __str__? Now all objects with the same host will be printed equally.

mreg/api/v1/zonefile.py
dmykarp: You mean it can be just
'fingerpring': fingerprint ?

mreg/models.py
dmykarp: Sorry, don't get it. the function name should be str?

mreg/api/v1/zonefile.py
oyvindkolbu: Yes

mreg/models.py
oyvindkolbu: Wrong usage of github-markup. Should say better `__str__`. Something like:
`return "f{self.host.name} -> {self.algorithm} ({self.get_algorithm_display()}) {self.hash_type} ({self.get_hash_type_display()}) {self.fingerprint}"`

Though wrap/make it shorter.

PR #283, opened 2019-04-03
Fine grained authorization

### LINKED ISSUE #67: Autorisering
- Det eneste vi trenger i første omgang er DHCP-delegeringen.
- Hvis vi kan se på medlemskap i nettgrupper som vi får fra LDAP så er det praktisk.  
- Et subnett eies ikke nødvendigvis bare av en lokal-IT, det kan være flere miljøer/grupper som trenger tilgang.

Oppgaver:  
- [ ] DHCP-delegering til deler av subnett (basert på cidr)
- [ ] Delegere tilgang til å sette hostnavn basert på prefiks

oyvindkolbu:
Adds reveral new classes to `mreg.api.permission`, most notably `IsGrantedNetGroupRegexPermission`. This class provides fine grained authorization through the ` NetGroupRegexPermission` model.

Unlike plain DRF which uses `has_object_permission`, this code has custom hooks for create/update/destroy methods. The main benefit is that the methods are called after the serializer has validated data, so no need the check that.

Lacks tests for testing reduced priveleges, but will come in a later pull.

Resolves #67.


mreg/api/permissions.py
```diff
@@ -10,7 +56,142 @@ class IsInRequiredGroup(BasePermission):
...
+    """
+    Permit user if in super or admin group, else read only.
+    """
+
+    def has_permission(self, request, view):
+        if not bool(request.user and request.user.is_authenticated):
+            return False
+        if request.method in SAFE_METHODS:
+            return True
+        self.user_groups = set(request.user.groups.values_list('name', flat=True))
```
ponas: Considering moving this to a convenience property on the user object

mreg/api/permissions.py
```diff
@@ -10,7 +56,142 @@ class IsInRequiredGroup(BasePermission):
...
+    def has_permission(self, request, view):
+        # This method is called before the view is executed, so
+        # just do some preliminary checks.
+        if not bool(request.user and request.user.is_authenticated):
+            return False
+        if request.method in SAFE_METHODS:
+            return True
+        if self.is_super_or_admin(request):
+            return True
+        if not user_in_required_group(self.user_groups):
```
ponas: Here, `self.user_groups` is only set as a side effect of calling `self.is_super_or_admin`

mreg/api/permissions.py
```diff
@@ -10,7 +56,142 @@ class IsInRequiredGroup(BasePermission):
...
+            # If renaming a host, make sure the user has permission to both the
+            # new and and old hostname.
+            if 'name' in data:
+                if not self.has_perm(data['name'], ips):
+                    return False
+            return self.has_perm(hostname, ips)
+        elif hasattr(obj, 'host'):
+            return self.has_obj_perm(obj.host)
+        else:
+            raise exceptions.permissiondenied(f"unhandeled view: {view}")
```
ponas: `permissiondenied` -> `PermissionDenied`, `unhandeled` -> `unhandled`

mreg/api/permissions.py
```diff
@@ -10,7 +56,142 @@ class IsInRequiredGroup(BasePermission):
...
+    def has_destroy_permission(self, request, view, validated_serializer):
+        if self.is_super_or_admin(request):
+            return True
+        obj = view.get_object()
+        if isinstance(view, mreg.api.v1.views.HostDetail):
+            pass
+        elif hasattr(obj, 'host'):
+            obj = obj.host
+        else:
+            raise exceptions.permissiondenied(f"unhandeled view: {view}")
```
ponas: Casing + typo

mreg/api/permissions.py
oyvindkolbu: Maybe a new issue? Would cause a lot more churn..

mreg/api/permissions.py
oyvindkolbu: Intentional, but probably not wise.
Can rework to setting it explicitly.

PR #301, opened 2019-04-12
Added tests for automatic operations on PtrOverride

dmykarp:


mreg/api/v1/tests.py
```diff
@@ -608,6 +608,55 @@ def test_ptr_override_list(self):
...
+                     'ipaddress': '10.0.0.5'}
+        host2_data = {'name': 'ns4.example.org',
+                      'contact': 'mail@example.org',
+                      'ipaddress': '10.0.0.5'}
+        self.client.post('/hosts/', host_data)
+        ret = self.client.get("/ptroverrides/")
+        new_count = ret.data['count']
+        self.assertEqual(new_count, old_count)
+        self.client.post('/hosts/', host2_data)
+        ret = self.client.get("/ptroverrides/")
```
oyvindkolbu: Also assert that the the hostname/hostid for 10.0.0.5 in `ret['results]`  is `ns3.example.org`

mreg/api/v1/tests.py
```diff
@@ -608,6 +608,55 @@ def test_ptr_override_list(self):
...
+        # Adding a new host with already existing IP should
+        # create a PtrOverride for it
+        ret = self.client.get("/ptroverrides/")
+        old_count = ret.data['count']
+        host_data = {'name': 'ns3.example.org',
+                     'contact': 'mail@example.org',
+                     'ipaddress': '10.0.0.5'}
+        host2_data = {'name': 'ns4.example.org',
+                      'contact': 'mail@example.org',
+                      'ipaddress': '10.0.0.5'}
```
oyvindkolbu: Why `nsX`?

dmykarp: To be in a line with ns1 and ns2, that are used in setUp.

mreg/api/v1/tests.py
dmykarp: Implemented now in 83d071a3ec2ca21fca2d22bd23dea7c46b74276b

PR #336, opened 2019-06-24
Issue 206 cname check

dmykarp:
The first part: includes check when creating/updating Hosts and Naptrs through the API.

A bit unsure if cname_conflict subroutine can be left as is in `view.py` file itself, or moved to `utils.py`. Please comment.

### DIFF
--- a/mreg/api/v1/views.py
+++ b/mreg/api/v1/views.py
@@ -43,6 +43,15 @@ from .serializers import (CnameSerializer, ForwardZoneDelegationSerializer,
 from .zonefile import ZoneFile
 
 
+def cname_conflict(cname):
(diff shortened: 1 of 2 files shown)


mreg/api/v1/views.py
```diff
@@ -43,6 +43,15 @@
 from .zonefile import ZoneFile
 
 
+def cname_conflict(cname):
+    try:
+        conflicting_cname = Cname.objects.get(name=cname)
```
oyvindkolbu: `conflicting_cname` is not needed.

dmykarp: True, removed now.

PR #351, opened 2019-11-04
Host and hostgroup history

oyvindkolbu:


mreg/models.py
```diff
@@ -663,6 +663,24 @@ def find_perm(cls, groups, hostname, ips):
...
 
 
+class History(models.Model):
+    """
+    Store history for various models.
+
+    Use the resource field to set the scope for each group of events.
+    """
+    user = models.CharField(max_length=64)
+    change_type = models.CharField(max_length=64)
```
oyvindkolbu: Currenly this is use e.g. as `Host¤create`, should it instead be split in two field? `model` and `action` ?

PR #360, opened 2019-11-13
Add network endpoint to get random unused ip

### LINKED ISSUE #343: API to get random free ip
Some users want to assign a random free IP instead of getting the first unused IP for a network to make network scanning harder. Especially for ipv6-networks this can be useful.

paalbra:
Fix #343

### DIFF
--- a/mreg/models.py
+++ b/mreg/models.py
@@ -1,4 +1,5 @@
 import ipaddress
+import random
 from collections import defaultdict
 from datetime import timedelta
 from functools import reduce
@@ -546,6 +547,16 @@ class Network(BaseModel):
                 continue
             if ip not in used:
                 return str(ip)
+        return None
+
+    def get_random_unused(self):
+        """
+        Return a random unused IP, if any.
+        """
+
+        unused = self.get_unused_ipaddresses()
+        if unused:
+            return str(random.choice(tuple(unused)))
         return None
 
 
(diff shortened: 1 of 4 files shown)


mreg/models.py
```diff
@@ -548,6 +549,16 @@ def get_first_unused(self):
...
         return None
 
+    def get_random_unused(self):
+        """
+        Return a random unused IP, if any.
+        """
+
+        unused = self.get_unused_ipaddresses()
+        if unused:
+            return str(random.choice(tuple(unused)))
```
oyvindkolbu: Not random enough for ipv6, as `get_unused_ipaddresses()` will only return the first 4000 addresses. A normal ipv6 network has 2**64 ipaddresses and the random function should use the entire address space.
https://github.com/unioslo/mreg/blob/3a595a8d67fc08ee17b9b3bc810b94d7913445ce/mreg/models.py#L514-L523

mreg/models.py
```diff
@@ -512,20 +515,24 @@ def get_unused_ipaddresses(self):
...
         network_ips = []
-        if isinstance(self.network, ipaddress.IPv6Network):
+        unusable = self.get_unusable_ipaddresses()
+        used = self.get_used_ipaddresses()
+        not_available = unusable | used
+        if self.network.num_addresses > MAX_UNUSED_LIST:
             # Getting all availible IPs for a ipv6 prefix can easily cause
             # the webserver to hang due to lots and lots of IPs. Instead limit
-            # to the first 4000 hosts. Should probably be configurable.
+            # to the first 4000 hosts.
```
paalbra: Comment still mentions the static 4000

PR #382, opened 2019-12-17
Speed up the reversezone by not creating objects, but rely on values_list()

oyvindkolbu:
No need to create the ipaddress, ptroverride and host objects when we only want
a few values.

Shaves off 2/3 of the time spent.


paalbra:
Maybe add some comments to the `get_ipaddresses` function while at it?

It's not completely obvious at first glance what kind of data is handled (the `(ip, ttl, hostname)` and `(ttl, hostname)` tuples)

PR #383, opened 2019-12-17
Add PtrOverrides to used/unused/counts etc for Network.

### LINKED ISSUE #361: get_used_ipaddress_count should include ptroverrides
Either a new api to indicate that you get both Ipaddress and PtrOverrides usage, or the current one should be changed to include addresses used by PtrOverride.

oyvindkolbu:
Sometimes ipaddresses can be used by a PtrOverride object without
one or more Ipaddress objects for the same ipaddress, so cope with that.

Resolves #361.

### DIFF
--- a/mreg/api/v1/serializers.py
+++ b/mreg/api/v1/serializers.py
@@ -113,8 +113,8 @@ class IpaddressSerializer(ValidationMixin, serializers.ModelSerializer):
                 # Allow mac to be bound to both an ipv4 and ipv6 address on the same vlan
                 if ipversion != network.network.version:
                     continue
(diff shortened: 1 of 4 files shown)


paalbra:
How well does this handle large IPv6 networks with many unused/used addresses?

And maybe add some tests for the case mentioned in the issue (if this isn't catched already)?

oyvindkolbu:
> How well does this handle large IPv6 networks with many unused/used addresses?

Yes.

> And maybe add some tests for the case mentioned in the issue (if this isn't catched already)?

Will do!

Issue #349, opened 2019-10-25
Update zone on delegation changes


oyvindkolbu:
After allowing patch of delegations, it should trigger an update of the zone's `updated_at` field on changes.

PR #397, opened 2020-02-04
Update zone for existing hosts when adding a zone

### LINKED ISSUE #393: Update zone for existing hosts when adding og removing zone
On create/rename/delete of hosts, update zone for existing hosts. Something like
```python
for h in Host.objects.filter(name__endswith=zone.name):
     z = ForwardZone.get_zone_by_hostname(h.name)
     h.zone = z
     h.save()
```

oyvindhagberg:
Also, update the zone timestamp when a delegation is patched.


paalbra:
Why is the parent zone updated? Maybe move this to another issue if it's an issue?

Test descriptions should probalby be docstrings.

oyvindhagberg:
> Why is the parent zone updated? Maybe move this to another issue if it's an issue?

This PR resolves both #393 and #349. If you'd rather have a separate branch and PR for each issue, we can do that.

> Test descriptions should probably be docstrings.

I will change them.

Issue #404, opened 2020-03-10
token-auth responds with 400 Bad Request if credentials are invalid


paalbra:
If you send invalid/unauthorized credentials to `/api/token-auth/` you will get a 400 Bad Request.

401 Unauthorized makes more sense.

PR #407, opened 2020-04-22
Make /api/token-auth respond 401 if invalid username/password

oyvindhagberg:


paalbra:
Changing the topic. This refers to the issue number #404 . This should not be confused with HTTP 404.

PR #408, opened 2020-05-04
Hosts in a delegation get zone = None, not the parent zone.

### LINKED ISSUE #362: Hosts in a delegation shall get zone == None, not the parentzone
E.g with zone `example.org` and delegation `sub.example.org` when host `hostname.sub.example.org` is created the `zone` attribute must be `None` and not point to  `example.org`.

Make sure that glue records for delegations still are included in the parentzone and are now at the top, where they should be.

oyvindhagberg:
Fixes #362.

### DIFF
--- a/mreg/api/v1/tests/tests_zones.py
+++ b/mreg/api/v1/tests/tests_zones.py
@@ -366,6 +366,20 @@ class ZonesForwardDelegationTestCase(MregAPITestCase):
         # the timestamp should have been updated
         self.assertLess(before, after)
 
+    def test_hosts_in_delegation_get_zone_none(self):
+        """ Hosts in a delegation shall get zone == None, not the parentzone. #362 """
+        # add a delegation
+        path = self.del_path('example.org')
+        data = {'name': 'delegated.example.org',
+                'nameservers': ['ns1.example.org', 'ns1.delegated.example.org']}
+        self.assert_post(path, data)
+        # add a host in the delegation
+        self.assert_post_and_201('/hosts/', {"name": "foo.delegated.example.org",
+                      "ipaddress": "10.10.0.1", "contact": "mail@delegated.example.org"})
+        # load the host object and verify that its zone is None
+        host = Host.objects.get(name="foo.delegated.example.org")
+        self.assertTrue(host.zone is None)
+
 
 class ZonesReverseDelegationTestCase(MregAPITestCase):
     """ This class defines test testsuite for api/zones/reverse/<name>/delegations/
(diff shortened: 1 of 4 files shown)


mreg/signals.py
```diff
@@ -298,3 +299,13 @@ def update_hosts_when_zone_is_added(sender, instance, created, **kwargs):
                 continue
             h.zone = instance
             h.save()
+
+@receiver(pre_save, sender=Host)
+def hosts_in_a_delegation_shall_get_zone_None(sender, instance, raw, using, update_fields, **kwargs):
+    """ Hosts in a delegation shall get zone == None, not the parentzone. #362 """
+    if instance.zone != None:
+        for d in ForwardZoneDelegation.objects.filter(zone = instance.zone):
+            if instance.name.endswith('.'+d.name):
```
oyvindkolbu, line 308: Does not handle an instance with the same name as the delegation.

mreg/signals.py
```diff
@@ -298,3 +299,13 @@ def update_hosts_when_zone_is_added(sender, instance, created, **kwargs):
                 continue
             h.zone = instance
             h.save()
+
+@receiver(pre_save, sender=Host)
+def hosts_in_a_delegation_shall_get_zone_None(sender, instance, raw, using, update_fields, **kwargs):
```
oyvindkolbu, line 304: Will slow down things massively. Only needed for new instances or updated name, though not sure if it should be done as a dedicated signal at all.

mbakke (approved):
:shipit:

oyvindkolbu (approved):
:shipit:

mreg/api/v1/tests/tests_zones.py
```diff
@@ -366,6 +366,26 @@ def test_delegation_patch_modifies_zone_updated_at(self):
...
+                "ipaddress": "10.10.0.1", "contact": "mail@delegated.example.org"})
+        # load the host object and verify that its zone is None
+        host = Host.objects.get(name="foo.delegated.example.org")
+        self.assertTrue(host.zone is None)
+        # add a host with the same name as the delegation
+        self.assert_post_and_201('/hosts/', {"name": "delegated.example.org",
+                "ipaddress": "10.10.0.2", "contact": "mail@delegated.example.org"})
+        # load the host object and verify that its zone is None
+        host = Host.objects.get(name="delegated.example.org")
+        self.assertTrue(host.zone is None)
```
oyvindkolbu, line 387: This is prettier handled with a `self.assertIsNone(host.zone)`.

Also multiple variantions of a test within a test function should use subtesting. See
https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

oyvindhagberg: Just added a commit (bb326f97c92ce3ed8ca1c09c8d5102606a551ab7) with these modifications.

mbakke (approved):

PR #411, opened 2020-09-07
Migrate from Swagger to OpenAPI

mbakke:
OpenAPI support is built-in to newer versions of djangorestframework, and coreapi/swagger is deprecated.

### DIFF
--- a/mregsite/urls.py
+++ b/mregsite/urls.py
@@ -3,10 +3,10 @@ from django.urls import include, path
 
 from mreg.api.v1 import views
 
-from rest_framework_swagger.views import get_swagger_view
+from rest_framework.schemas import get_schema_view
 
 # Schema view for swagger api documentation
-schema_view = get_swagger_view(title='mreg API')
+schema_view = get_schema_view(title='mreg API')
 
 urlpatterns = [
     path('api/', include('mreg.api.urls')),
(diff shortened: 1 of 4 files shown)


mregsite/urls.py
```diff
@@ -3,10 +3,12 @@
...
 from mreg.api.v1 import views
 
-from rest_framework_swagger.views import get_swagger_view
+from rest_framework.schemas import get_schema_view
+from rest_framework.permissions import AllowAny
 
 # Schema view for swagger api documentation
-schema_view = get_swagger_view(title='mreg API')
+schema_view = get_schema_view(title='mreg API',
+                              permission_classes=[AllowAny])
```
oyvindkolbu, line 11: Maybe add a test that the docs are readable without any authentication?

oyvindkolbu (approved):
:shipit: a test for the docs would be nice

ponas (dismissed):

mregsite/urls.py
mbakke: @oyvindkolbu Thanks for reviewing.  I decided to drop the second commit entirely for now.  AIUI the `docs/` endpoint should provide a full API overview, but as of currently it only documents itself (i.e. `GET /docs`, not very useful).  That behaviour is the same before and after this PR.

I'll look into adjusting `/docs` to provide an actual OpenAPI UI in a separate PR.

mbakke:
v2 of this PR drops the mostly-unrelated `/docs` endpoint change, and removes stale references to `django-rest-swagger`.

oyvindhagberg (approved):

PR #413, opened 2020-09-08
Update to Django 3.1

mbakke:
Update Django and dependencies to the latest available versions, adjust `JSONField()` import for Django 3.1.

While at it, update dependencies and remove the unused(?) `sentry-sdk`.


requirements.txt
```diff
@@ -1,10 +1,9 @@
...
-django-auth-ldap==2.1.0
+django-auth-ldap==2.2.0
 django-logging-json==1.15
 django-netfields==1.2.2
-django-url-filter==0.3.14
+django-url-filter==0.3.15
 gunicorn==20.0.4
-idna==2.8
-psycopg2-binary==2.8.4
-sentry-sdk==0.14.0
```
ponas, line 10: `sentry-sdk` is used in the test and production environments to collect errors and tracebacks.

ponas (changes requested):

mreg/models.py
```diff
@@ -769,7 +769,7 @@ class History(models.Model):
     model_id = models.PositiveIntegerField()
     model = models.CharField(max_length=64)
     action = models.CharField(max_length=64)
-    data = pgfields.JSONField()
+    data = models.JSONField()
```
ponas, line 772: Changing the field requires you to [make a migration](https://docs.djangoproject.com/en/3.1/ref/django-admin/#django-admin-makemigrations)

ponas (changes requested):

requirements.txt
oyvindkolbu, line 10: sentry-sdk is used by `local_settings.py` to send messages to sentry instances.

mreg/models.py
mbakke: Good catch; I had already made it but forgot to commit it.

requirements.txt
mbakke: Ah, I see. Dropped the commit and instead bumped sentry-sdk in patch 2/2.  Thanks!

mreg/migrations/0007_auto_20200905_1410.py
```diff
@@ -0,0 +1,23 @@
...
+        migrations.AlterField(
+            model_name='history',
+            name='data',
+            field=models.JSONField(),
+        ),
+        migrations.AlterField(
+            model_name='user',
+            name='first_name',
+            field=models.CharField(blank=True, max_length=150, verbose_name='first name'),
+        ),
```
mbakke, line 22: The user migration comes from https://docs.djangoproject.com/en/3.1/releases/3.1/#abstractuser-first-name-max-length-increased-to-150

ponas (approved):

PR #414, opened 2020-09-09
Allow MAC address to be specified in various formats

### LINKED ISSUE #394: Consider allowing other mac address formats as input
```
PATCH /api/v1/ipaddresses/1
{
    "host": 1,
    "macaddress": "AA:BB:CC:00:11:22"
}
```

```
HTTP 400 Bad Request
{
    "macaddress": [
        "Must be on form: aa:bb:cc:00:11:22"
    ]
}
```

Other possible formats: "-" separator, only aphanumeric (upper/lower).

mbakke:
Allow MAC address to be specified using `:.-` as optional separators, and also allow uppercase letters.

Note that this will allow MACs such as `AA:BB.cc-ddeeff` and normalize it to `aa:bb:cc:dd:ee:ff`.  Not sure if that is too lenient.

Fixes #394.

### DIFF
--- a/mreg/validators.py
+++ b/mreg/validators.py
@@ -122,6 +122,12 @@ def validate_mac_address(address):
                                message="Must be on form: aa:bb:cc:00:11:22")
     validator(address)
 
+def validate_normalizeable_mac_address(address):
+    """Validates that the mac address can be normalized."""
+    adr_regex = "^([a-fA-F0-9]{2}[:.-]?){5}[a-fA-F0-9]{2}$"
+    validator = RegexValidator(adr_regex,
+                               message="Invalid MAC address.")
+    validator(address)
 
 def validate_loc(location):
     """Validates that the loc input is on a valid form."""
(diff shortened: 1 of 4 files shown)


mreg/validators.py
```diff
@@ -117,9 +117,9 @@ def validate_reverse_zone_name(name):
 
 def validate_mac_address(address):
     """Validates that the mac address is on a valid form."""
-    adr_regex = "^([a-f0-9]{2}:){5}[a-f0-9]{2}$"
+    adr_regex = "^([a-fA-F0-9]{2}[:.-]?){5}[a-fA-F0-9]{2}$"
     validator = RegexValidator(adr_regex,
-                               message="Must be on form: aa:bb:cc:00:11:22")
+                               message="Not a valid MAC address")
```
oyvindkolbu, line 122: Any reason why you changed the validator? This would make it possible to actually store "aa:AA:aa:bb:ccAA" as a mac-address if created outside the serializer, e.g. signals or scripts.

mbakke: Good point.  I've pushed a v2 that implements this check only in the serializer, and separates the conversion out to a [SerializerMethodField](https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield).

oyvindkolbu (approved):
:shipit:

ponas (approved):

PR #416, opened 2020-09-10
Allow patching IP with existing MAC address.

mbakke:
This addresses <https://github.com/unioslo/mreg-cli/issues/131>.

The intention is to disable MAC address uniqueness validation when the MAC does not change, or when it belongs to the IP being patched.


oyvindkolbu (approved):
:shipit:

oyvindhagberg:
👍

PR #418, opened 2020-09-22
Prevent adding hosts that conflict with CNAME records.

mbakke:
See #417 for background.

### DIFF
--- a/mreg/api/errors.py
+++ b/mreg/api/errors.py
@@ -1,0 +1,4 @@
+from rest_framework import (exceptions, status)
+
+class ValidationError409(exceptions.APIException):
+    status_code = status.HTTP_409_CONFLICT
--- a/mreg/api/v1/serializers.py
+++ b/mreg/api/v1/serializers.py
@@ -13,6 +13,7 @@ from mreg.models import (Cname, ForwardZone, ForwardZoneDelegation,
                          ReverseZone, ReverseZoneDelegation, Srv, Sshfp, Txt)
 from mreg.utils import (nonify, normalize_mac)
 from mreg.validators import (validate_keys, validate_normalizeable_mac_address)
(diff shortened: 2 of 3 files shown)


mbakke:
That's a great point, confirmed by a new test:

```
    def test_hosts_patch_409_conflict_cname(self):
        """Patching an entry with a CNAME that already exists should return 409"""
        post_data_cname = {'name': 'new-name2.example.org',
                           'host': self.host_two.id}
        self.assert_post('/cnames/', post_data_cname)
        self.assert_patch_and_409('/hosts/%s' % self.host_one.name,
                                  {'name': 'new-name2.example.org'})
```

I suppose the CNAME check should be done in serializer, will investigate how to raise a specific HTTP status code as discussed in #417.

mbakke:
@oyvindkolbu I've pushed a v2 that adds a new `ValidationError409` class and raises it in the serializer on CNAME conflicts.

Not sure whether `serializers.py` is a good home for that `ValidationError409` class though.

mbakke:
v3 of this patch introduces `mreg.errors` for the custom `ValidationError409` class.

oyvindhagberg (approved):

mreg/errors.py
```diff
@@ -0,0 +1,4 @@
+from rest_framework import (exceptions, status)
+
+class ValidationError409(exceptions.APIException):
+    status_code = status.HTTP_409_CONFLICT
```
oyvindkolbu, line 4: Except that this file should probably reside under `mreg/api/` or `mreg/api/v1/` it looks good.

mbakke: That makes sense.  Moved to `mreg.api.errors`.  Thanks a lot for the feedback!  :-)

oyvindhagberg (approved):

PR #420, opened 2020-10-19
Change requirements for mac address uniqueness and fix bugs

oyvindhagberg:
While investigating [RT #4088062](https://rt.uio.no/Ticket/Display.html?id=4088062), I found a couple of bugs.
I also suggest removing the requirement that mac addresses should be unique across VLANs.
See the commits for details.

Btw, to solve the RT issue, it looks like we may also need to make changes and/or fix bugs in mreg-cli.

### DIFF
--- a/mreg/api/v1/serializers.py
+++ b/mreg/api/v1/serializers.py
@@ -95,9 +95,7 @@ class IpaddressSerializer(ValidationMixin, serializers.ModelSerializer):
             # by multiple IP addresses, although normally it would never be more than 1.
             inuse_set = qs.filter(macaddress=mac)
             if inuse_set.exists():
-                ips = []
-                for inuse_ip in inuse_set:
-                    ips.append("{}".format(inuse_ip.ipaddress))
+                ips = inuse_set.values_list('ipaddress', flat=True)
                 msg = "macaddress already in use by: " + ", ".join(ips)
                 raise ValidationError409(msg)
 
--- a/mreg/api/v1/serializers.py
+++ b/mreg/api/v1/serializers.py
@@ -94,10 +94,15 @@ class IpaddressSerializer(ValidationMixin, serializers.ModelSerializer):
         """
 
         def _raise_if_mac_found(qs, mac):
-            if qs.filter(macaddress=mac).exists():
-                inuse_ip = qs.get(macaddress=mac).ipaddress
-                raise serializers.ValidationError(
-                    "macaddress already in use by {}".format(inuse_ip))
+            # There is a theoretical possibility that a mac address can be in use
+            # by multiple IP addresses, although normally it would never be more than 1.
+            inuse_set = qs.filter(macaddress=mac)
+            if inuse_set.exists():
+                ips = []
+                for inuse_ip in inuse_set:
+                    ips.append("{}".format(inuse_ip.ipaddress))
+                msg = "macaddress already in use by: " + ", ".join(ips)
+                raise ValidationError409(msg)
 
         data = super().validate(data)
         _validate_ip_not_in_network_excluded_range(data.get('ipaddress'))
--- a/mreg/api/v1/serializers.py
+++ b/mreg/api/v1/serializers.py
@@ -134,9 +134,9 @@ class IpaddressSerializer(ValidationMixin, serializers.ModelSerializer):
                 if ipversion != network.network.version:
                     continue
                 qs = network._used_ipaddresses()
-                # Validate the MAC unless it belonged to the old IP.
-                if self.instance and not self.instance in qs:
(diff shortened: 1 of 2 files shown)


oyvindkolbu:
> While investigating [RT #4088062](https://rt.uio.no/Ticket/Display.html?id=4088062), I found a couple of bugs.
> I also suggest removing the requirement that mac addresses should be unique across VLANs.
> See the commits for details.
> 
> Btw, to solve the RT issue, it looks like we may also need to make changes and/or fix bugs in mreg-cli.

I don't have access to the RT issue, so not sure what the issue is.

oyvindhagberg:
> > While investigating [RT #4088062](https://rt.uio.no/Ticket/Display.html?id=4088062), I found a couple of bugs.
> > I also suggest removing the requirement that mac addresses should be unique across VLANs.
> > See the commits for details.
> > Btw, to solve the RT issue, it looks like we may also need to make changes and/or fix bugs in mreg-cli.
> 
> I don't have access to the RT issue, so not sure what the issue is.

Andreas hadde problemer med å assosiere en mac-adresse til to forskjellige ip-adresser under migrering mellom nett.
Det viste seg å skyldes at nettene var på samme VLAN, men det ble også avdekket noe rar oppførsel i mreg-cli. Vi kommer tilbake til det. Når det gjelder selve Mreg ble det oppdaget et par småting som bør rettes, det er det som er i denne PR'en, i tillegg til at den tar bort begrensningen med mac-adresser og VLAN.

mreg/api/v1/serializers.py
```diff
@@ -85,19 +85,21 @@ class Meta:
...
-                inuse_ip = qs.get(macaddress=mac).ipaddress
-                raise serializers.ValidationError(
-                    "macaddress already in use by {}".format(inuse_ip))
+            # There is a theoretical possibility that a mac address can be in use
+            # by multiple IP addresses, although normally it would never be more than 1.
+            inuse_set = qs.filter(macaddress=mac)
+            if inuse_set.exists():
+                ips = []
+                for inuse_ip in inuse_set:
+                    ips.append("{}".format(inuse_ip.ipaddress))
```
oyvindkolbu, line 100: Better:
```
ips = inuse_set.values_list('ipaddress', flat=True)
```

oyvindkolbu (changes requested):
Still think the old restriction of only one mac in use for an ipv4 and ipv6 network in a vlan holds, but I can be wrong.

mreg/api/v1/serializers.py
oyvindhagberg: Made a commit just now with your suggested change.

As for the vlan restriction, there's a test for it that looks correct to me:
https://github.com/unioslo/mreg/blob/7af8098a4a468733e3bee2516bc51ab82a613395/mreg/api/v1/tests/tests.py#L1288-L1310

I also tested it with mreg-cli just now. Ended up with this:
```
mreg> host info eple
Name:         eple.uio.no
Contact:
A_Records     IP              MAC
              10.0.0.5        aa:bb:cc:dd:ee:ff
TTL:          (Default)
mreg> host info banan
Name:         banan.uio.no
Contact:
AAAA_Records  IP              MAC
              2001:db8:1::4   aa:bb:cc:dd:ee:ff
TTL:          (Default)
mreg> network info 10.0.0.0/24
Network:                 10.0.0.0/24
Netmask:                 255.255.255.0
Description:             test1
Category:
Location:
VLAN                     1
DNS delegated:           False
Frozen                   False
IP-range:                10.0.0.0 - 10.0.0.255
Reserved host addresses: 3
                         10.0.0.0 (net)
                         10.0.0.1
                         10.0.0.2
                         10.0.0.3
                         10.0.0.255 (broadcast)
Used addresses:          1
Unused addresses:        250 (excluding reserved adr.)
mreg> network info 2001:db8:1::/64
Network:                 2001:db8:1::/64
Netmask:                 ffff:ffff:ffff:ffff:0000:0000:0000:0000
Description:             hei og hå
Category:
Location:
VLAN                     1
DNS delegated:           False
Frozen                   False
IP-range:                2001:db8:1:: - 2001:db8:1:0:ffff:ffff:ffff:ffff
Reserved host addresses: 3
                         2001:db8:1:: (net)
                         2001:db8:1::1
                         2001:db8:1::2
                         2001:db8:1::3
Used addresses:          1
Unused addresses:        18446744073709551611 (excluding reserved adr.)
mreg>
```

oyvindkolbu (approved):
Reluctantly approving the change. Still the old constraint holds up.

mbakke (approved):
All LGTM.  I can confirm that the `mreg-cli` examples from the RT issue all work with this PR.

PR #423, opened 2020-12-04
Feature hostpolicy labels

oyvindhagberg:
Solution to #370

* This PR adds a new data class, Label, and it has many-to-many-relationships with HostPolicyRole and NetGroupRegexPermission.
* The access control for adding roles to hosts now allows regular users to do so, if there's a matching label between the role and permission.


hostpolicy/api/permissions.py
```diff
@@ -13,15 +16,58 @@ def is_super_or_hostpolicy_admin(user):
...
+                if label.name in role_labels:
+                    return True
+
+        # If the code got to this point, it means none of the labels matched.
+        return False
+
 
     def has_m2m_change_permission(self, request, view):
-        return is_super_or_hostpolicy_admin(request.user)
+        return True
```
ponas, line 73: Seems to permissive, but might be fine?

ponas (approved):

hostpolicy/api/permissions.py
oyvindkolbu: Pretty sure this OK. It is just an early check to see if the request can rejected before doing the heavy lifting.

hostpolicy/api/permissions.py
```diff
@@ -13,15 +16,58 @@ def is_super_or_hostpolicy_admin(user):
...
+        # the ipaddress, hostname, and the groups that the user is a member of
+        if 'host' in view.kwargs:
+            hostname = view.kwargs['host']
+        else:
+            hostname = request.data.get("name")
+        host = Host.objects.get(name=hostname)
+        ips = []
+        hostserr = HostSerializer(host)
+        for i in hostserr.data['ipaddresses']:
+            ips.append(i['ipaddress'])
```
oyvindkolbu, line 54: ``ips = Host.objects.filter(name=hostname).values_list('ipaddresses__ipaddress', flat=True)``

hostpolicy/api/permissions.py
```diff
@@ -13,15 +16,58 @@ def is_super_or_hostpolicy_admin(user):
...
+        # In that case, non-admin-users shouldn't have access anyway, and we can deny the request.
+        if not (view.__class__.__name__ == 'HostPolicyRoleHostsDetail' or
+                view.__class__.__name__ == 'HostPolicyRoleHostsList'):
+            return False
+
+        # Find out which labels are attached to this role
+        role = HostPolicyRole.objects.get(name=view.kwargs['name'])
+        role_labels = []
+        for label in role.labels.all():
+            role_labels.append(label.name)
```
oyvindkolbu, line 42: ``role_labels =  HostPolicyRole.objects.filter(name=view.kwargs['name']).values_list('labels__name')``

hostpolicy/api/permissions.py
```diff
@@ -13,15 +16,58 @@ def is_super_or_hostpolicy_admin(user):
...
+        # If no permissions matched the host/ip, we deny access
+        if not qs.exists():
+            return False
+
+        # Do any of those permissions have labels that match the labels attached to this role?
+        # If so, access is granted
+        for perm in qs:
+            for label in perm.labels.all():
+                if label.name in role_labels:
+                    return True
```
oyvindkolbu, line 66:
```
perm_labels = qs.values_list('labels__name')
if any(label in perm_labels for label in role_labels):
   return True
``

File: manage.py [hunk 5]
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
oyvindkolbu, line 1: No. This is handled by the python environment.

mreg/api/v1/views_labels.py
```diff
@@ -0,0 +1,62 @@
...
+        if "name" in request.data:
+            if self.get_queryset().filter(name=request.data['name']).exists():
+                content = {'ERROR': 'Label name already in use'}
+                return Response(content, status=status.HTTP_409_CONFLICT)
+            if ' ' in request.data['name']:
+                content = {'ERROR': "Label name can't contain spaces"}
+                return Response(content, status=status.HTTP_400_BAD_REQUEST)
+            if not "description" in request.data or request.data['description'] == '':
+                content = {'ERROR': "A description is required for a label"}
+                return Response(content, status=status.HTTP_400_BAD_REQUEST)
```
oyvindkolbu, line 37: The last two should be handled by the model. Otherwise one can use `PATCH/UPDATE`  to rename to invalid names or descriptions.

mreg/models.py
```diff
@@ -718,10 +718,22 @@ def __str__(self):
         return "%s" % self.name
 
 
+class Label(BaseModel):
+    name = LCICharField(max_length=64, unique=True)
+    description = models.TextField(blank=True)
```
oyvindkolbu, line 723: Add restrictions here via `validators`.

oyvindkolbu (changes requested):
Needs some improvements.

hostpolicy/api/permissions.py
paalbra: I'm not familiar with these parts. Couldn't the function call just be removed? Why have a permission that always return True?

oyvindkolbu: Could have been if there was a parent class which provided these by default, but this class just inherits `IsAuthenticated` and all the `m2m`-stuff is mreg-only code.
The usage is https://github.com/unioslo/mreg/blob/feature-hostpolicy-labels/mreg/api/v1/views_m2m.py#L14
It is modeled similar to Django's non-m2m objects.

hostpolicy/api/permissions.py
oyvindkolbu: Maybe even a variation with:
```
return qs.filter(labels__name__in=role_labels).exists():  
```

hostpolicy/api/permissions.py
oyvindhagberg: Had to filter out some `None` values, ended up with this:
https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/hostpolicy/api/permissions.py#L47

hostpolicy/api/permissions.py
oyvindhagberg: Done:
https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/hostpolicy/api/permissions.py#L39

hostpolicy/api/permissions.py
oyvindhagberg: Done:
https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/hostpolicy/api/permissions.py#L56-L58

manage.py
oyvindhagberg: I changed it back now.
https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/manage.py#L1

mreg/api/v1/views_labels.py
oyvindhagberg: Removed the last two checks:
https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/mreg/api/v1/views_labels.py#L27-L33

mreg/models.py
oyvindhagberg: Added a validator to check for spaces in a string, and used the `blank` option to disallow empty values for the description:
https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/mreg/models.py#L722-L724

https://github.com/unioslo/mreg/blob/17eb02178d5f4025b27754ab3ce4def75e06fd8e/mreg/validators.py#L172-L174

oyvindkolbu (approved):

PR #427, opened 2021-02-23
Fix a bug/security hole in the hostpolicy API

oyvindhagberg:
A non-privileged user with network-group-regex permission to a host can add any roles to the host as long as the role does not have any labels.

### DIFF
--- a/hostpolicy/api/permissions.py
+++ b/hostpolicy/api/permissions.py
@@ -37,6 +37,9 @@ class IsSuperOrHostPolicyAdminOrReadOnly(IsAuthenticated):
 
         # Find out which labels are attached to this role
         role_labels = HostPolicyRole.objects.filter(name=view.kwargs['name']).values_list('labels__name',flat=True)
+        if not any(role_labels):
+            # if the role doesn't have any labels, there's no possibility of access at this point
+            return False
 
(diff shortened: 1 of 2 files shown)


hostpolicy/api/permissions.py
```diff
@@ -37,6 +37,9 @@ def has_permission(self, request, view):
 
         # Find out which labels are attached to this role
         role_labels = HostPolicyRole.objects.filter(name=view.kwargs['name']).values_list('labels__name',flat=True)
+        if not any(role_labels):
+            # if the role doesn't have any labels, there's no possibility of access at this point
+            return False
```
oyvindkolbu, line 42: Don't see how this should cause a security issue. Also a simple `if not role_labels` should suffice.

paalbra (approved):
The surrounding code could be cleaner/easier to read, but it resolves the issue

hostpolicy/api/permissions.py
paalbra: TLDR: `any(label in [None] for label in [None])` is the issue

oyvindkolbu: Ah, makes sense and not at the same time. Returning `[None]` instead of an empty list was a bit unexpected.

oyvindkolbu (approved):

PR #428, opened 2021-03-02
Lift 255 char limit for TXTs to an arbitrary limit.

### LINKED ISSUE #373: Accept TXT of more than 255 characters, if in < 255 character quoted groups
Lift the hard limit of 255 characters, and instead implement something like:
https://kb.isc.org/docs/aa-00356

oyvindkolbu:
If the TXT data is longer than 255, split it into chuncks of
255 characters with a space between them. This is in line with
RFC 4408 section 3.1.3.

Closes #373.


mreg/api/v1/zonefile.py
```diff
@@ -121,7 +121,12 @@ def sshfp_zf_string(self, name, ttl, algorithm, hash_type, fingerprint):
 
     def txt_zf_string(self, name, ttl, txt):
         record_type = 'TXT   '
-        record_data = quote_if_space(txt)
+        record_data = ''
+        # Support RFC 4408 section 3.1.3 style TXTs: long strings splitted in 255 character chunks
+        for i in range(0, len(txt), 255):
+            if i > 0:
+                record_data += ' '
+            record_data += quote_if_space(txt[i:i+255])
```
mbakke, line 129: I'm not sure whether adding a space every 256 characters is a good approach.  What if it splits a record in the middle of something important?

IMO it would be more "obvious" to stick closer to the RFC: split into a list of 255 character strings, and return those.  Then it's up to the client to concatenate them.  WDYT?

oyvindkolbu: > I'm not sure whether adding a space every 256 characters is a good approach. What if it splits a record in the middle of something important?

The entire content is important, not a single character or word.

> 
> IMO it would be more "obvious" to stick closer to the RFC: split into a list of 255 character strings, and return those. Then it's up to the client to concatenate them. WDYT?

I don't really see what you are suggesting is differnt from what I coded? A long string is splitted into a list of 255 characters strings.

The client is responsible for joining the list, as per the RFC:

> If a published record contains multiple strings, then the
   record MUST be treated as if those strings are concatenated together
   without adding spaces

oyvindhagberg (approved):
LGTM. This is probably correct. And if not, I'm sure we'll hear about it. 👍 :shipit:

PR #432, opened 2021-04-14
New feature: A special LDAP group allows members to create wildcard DNS records

oyvindhagberg:
This is a way to delegate creation of wildcard DNS records to non-super-users. Group membership in that LDAP group does not give any other additional forms of access.
Background: Some people on UiO need this to fully automate OpenShift cluster generation with Terraform.

### DIFF
--- a/mreg/api/permissions.py
+++ b/mreg/api/permissions.py
@@ -159,15 +159,17 @@ def _deny_superuser_only_names(data=None, name=None, view=None, request=None):
         if not name:
             if 'host' in data:
                 name = data['host'].name
-    if '*' in name:
-        # members of the dns wildcard group can create wildcard records below subdomains (*.sub.example.com)
-        if request_in_settings_group(request, DNS_WILDCARD_GROUP) and name.count('.') >= 3 and not '_' in name:
-            return False
-        return True
+
     # Underscore is allowed for non-superuser in SRV records
     if '_' in name and not isinstance(view, (mreg.api.v1.views.SrvDetail,
                                              mreg.api.v1.views.SrvList)):
         return True
+
+    # Except for super-users, only members of the DNS wildcard group can create wildcard records.
+    # And then only below subdomains, like *.sub.example.com
+    if '*' in name and (not request_in_settings_group(request, DNS_WILDCARD_GROUP) or name.count('.') < 3):
+        return True
+
     return False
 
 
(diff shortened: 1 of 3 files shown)


mreg/api/v1/tests/test_host_permissions.py
```diff
@@ -293,3 +294,15 @@ def test_super_only_delete_wildcard(self):
         path = '/api/v1/hosts/*.example.org'
         self.assert_delete_and_403(path)
         self.assert_delete(path, client=self.super_client)
+
+    def test_special_group_members_create_wildcard(self):
+        """Members in DNS_WILDCARD_GROUP can create entries with a wildcard, but only below subdomains"""
+        self.user = self.admin_user
```
oyvindkolbu, line 300: Can instead call:
`self.client = self.get_token_client(superuser=False, adminuser=False)` 
to get an client without extra access. Or is it intentional that you need to be both an admin _and_ `DNS_WILDCARD_GROUP`-member?

mreg/api/permissions.py
```diff
@@ -159,6 +160,9 @@ def _deny_superuser_only_names(data=None, name=None, view=None):
             if 'host' in data:
                 name = data['host'].name
     if '*' in name:
+        # members of the dns wildcard group can create wildcard records below subdomains (*.sub.example.com)
+        if request_in_settings_group(request, DNS_WILDCARD_GROUP) and name.count('.') >= 3 and not '_' in name:
+            return False
```
oyvindkolbu, line 165: Instead checking for underscore here, move the wildcard check after the under underscore in SRV test. Wildcards are already blocked in SRVs, so thats safe.

oyvindkolbu (changes requested):

mreg/api/v1/tests/test_host_permissions.py
oyvindhagberg: Yes. If the user isn't admin, it needs a NetgroupRegexPermission to be able to create a host. So the test function would have to add a permission for the user.

mreg/api/permissions.py
oyvindhagberg: Thanks for the tip. Implemented it in this new commit:
https://github.com/unioslo/mreg/pull/432/commits/e5b5827c30004cfa467ad40f9578307ea77387f6

oyvindkolbu (approved):
:shipit:

PR #436, opened 2021-05-06
Fix 435

oyvindhagberg:
Modified the functions that deal with unused/unusable IP addresses so they'll work with large excluded IPv6 ranges.

### DIFF
--- a/mreg/models.py
+++ b/mreg/models.py
@@ -518,20 +518,34 @@ class Network(BaseModel):
             ret.add(network.broadcast_address)
         return ret
 
-    def is_usable(self, ip) -> bool:
-        """ Checks whether the given ip address can be used.
-            Unusable addresses are:
-            - reserved addresses
-            - ips in excluded ranges
+    def get_unused_ipaddresses(self, max = MAX_UNUSED_LIST):
         """
-        if ip in self.get_reserved_ipaddresses():
-            return False
-        for i in self.excluded_ranges.all():
-            start_ip = ipaddress.ip_address(i.start_ip)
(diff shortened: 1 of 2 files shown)


mreg/models.py
```diff
@@ -574,43 +575,43 @@ def unused_addresses(self):
...
-        else:
-            return set(self.network.hosts()) - not_available
+        used = self.used_addresses
+        # Getting all availible IPs for a ipv6 prefix can easily cause
+        # the webserver to hang due to lots and lots of IPs. Instead limit
+        # to the first MAX_UNUSED_LIST hosts.
+        found = 0
+        for ip in self.network.hosts():
+            if ip in used or not self.is_usable(ip):
+                continue
```
oyvindkolbu, line 585: This will break in a similar way as previously if e.g. the exclude range is from `2001:db8::` to `2001:db8::ffff:0:0` just a tiny fraction on the `2001:db8::/64` network, but would still take time to iterate 2^48 addresses.

Maybe have some logic to check for first valid ip after an exclude range, instead of blindly attempting an ipaddress and then the next etc.

oyvindkolbu (changes requested):

oyvindhagberg:
So what do you think, is this good enough that we can approve and merge it or should we change anything?

oyvindkolbu:
> So what do you think, is this good enough that we can approve and merge it or should we change anything?

Update with the above improvement to use `for start_ip, end_ip...` to get proper variables instead of `ex[0]` and `ex[1]`. Technically it is OK.

oyvindkolbu (approved):

PR #441, opened 2021-06-02
When reserved adresses are in use, unused_count will subtract them twice

### LINKED ISSUE #440: When reserved adresses are in use, unused_count will subtract them twice
As long as the code in [models.py](https://github.com/unioslo/mreg/blob/master/mreg/models.py) is setup like this:
```python
# subtract reserved addresses
result -= len(self.get_reserved_ipaddresses())
# subtract used addresses
result -= len(self.used_addresses)
```
it will always count reserved IP addresses twice, when they are also in use, which in turn while lead to situations where for example mreg returns that there are `-2` unused IPs, because you have 3 reserved IPs (that are in use) and a single unused IP address.

oyvindhagberg:
Fixes #440.

### DIFF
--- a/mreg/api/v1/tests/test_networks.py
+++ b/mreg/api/v1/tests/test_networks.py
@@ -525,6 +525,9 @@ class NetworkExcludedRanges(MregAPITestCase):
     def test_count_unused_ips(self):
         """Test that excluded ipaddresses are exempted from the unused count"""
         self.test_create()
+        # Create a host that uses one of the reserved addresses. (Should not be counted twice)
+        host1 = Host.objects.create(name='host1.example.org')
+        Ipaddress.objects.create(host=host1, ipaddress='10.0.0.255')
         ret = self.assert_get(f'/api/v1/networks/{self.network4}/unused_count').json()
         # /24 - net/broadcast - reserved - [10, 200] - 201
         # 256 - 2             - 3        - 191       - 1   = 59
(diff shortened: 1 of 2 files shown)


mreg/models.py
```diff
@@ -609,10 +609,13 @@ def unused_count(self):
...
-        # subtract reserved addresses
-        result -= len(self.get_reserved_ipaddresses())
         # subtract used addresses
-        result -= len(self.used_addresses)
+        used = self.used_addresses
+        result -= len(used)
+        # subtract reserved addresses that aren't used
+        for r in self.get_reserved_ipaddresses():
+            if r not in used:
+                result-=1
```
oyvindkolbu, line 618: A bit complicated.
```
used_or_reserved = self.used_addresses | self.get_reserved_ipaddresses()
result -= len(used_or_reserved)
```

mreg/api/v1/tests/test_networks.py
```diff
@@ -525,6 +525,9 @@ def test_reject_ip_usage(self):
     def test_count_unused_ips(self):
         """Test that excluded ipaddresses are exempted from the unused count"""
         self.test_create()
+        # Create a host that uses one of the reserved addresses. (Should not be counted twice)
+        host1 = Host.objects.create(name='host1.example.org')
+        Ipaddress.objects.create(host=host1, ipaddress='10.0.0.255')
```
oyvindkolbu, line 530: Technically not one of the reserved address, as 10.0.0.255 is the broadcast address. The reserved addresses are the N (usually 3) first addresses, i.e. 10.0.0.{1,2,3}.

Could add another `Ipaddress` with e.g. 10.0.0.2 just to make sure it is counted properly.

oyvindkolbu (changes requested):

oyvindkolbu (approved):
:shipit:

PR #442, opened 2021-06-09
Rotate tokens every login

mbakke:
Hello,

This PR ensures that API tokens are rotated every login.

Using a token within the expiry window will extend the expiry time so e.g. automated jobs can keep using the same token.


mreg/authentication.py
```diff
@@ -23,4 +23,8 @@ def authenticate_credentials(self, key):
         if token.created < timezone.now() - timedelta(hours=EXPIRE_HOURS):
             raise exceptions.AuthenticationFailed('Token has expired')
 
+        # Update the created time of the token to keep it valid.
+        token.created = timezone.now()
```
ponas, line 27: If the token expiry check above used `token.updated` and that same field was updated here, you would be able to detect long-lived tokens.

ponas (approved):

mreg/authentication.py
oyvindkolbu: I'd say this is a good idea. Maybe use the `token.updated` field for extending the lifetime and keep `token.created` to actually be the time when it was created.

mbakke: I like this idea too.  One complication is that the `Token` model needs to be overridden.

I tried this approach:

```
diff --git a/mreg/api/views.py b/mreg/api/views.py
index 3834e62..adacb03 100644
--- a/mreg/api/views.py
+++ b/mreg/api/views.py
@@ -1,7 +1,6 @@
 from django.utils import timezone
 
 from rest_framework import status
-from rest_framework.authtoken.models import Token
 from rest_framework.authtoken.views import ObtainAuthToken
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
@@ -9,6 +8,8 @@ from rest_framework.views import APIView
 from rest_framework import serializers
 from rest_framework.exceptions import AuthenticationFailed
 
+from mreg.models import ExpiringToken
+
 
 class ObtainExpiringAuthToken(ObtainAuthToken):
 
@@ -26,9 +27,9 @@ class ObtainExpiringAuthToken(ObtainAuthToken):
         user = serializer.validated_data['user']
 
         # Force token rotation.
-        Token.objects.filter(user=user).delete()
+        ExpiringToken.objects.filter(user=user).delete()
 
-        token, created = Token.objects.get_or_create(user=user)
+        token, created = ExpiringToken.objects.get_or_create(user=user)
 
         return Response({'token': token.key})
 
diff --git a/mreg/authentication.py b/mreg/authentication.py
index cb97401..686d69f 100644
--- a/mreg/authentication.py
+++ b/mreg/authentication.py
@@ -7,24 +7,26 @@ from django.utils import timezone
 from rest_framework import exceptions
 from rest_framework.authentication import TokenAuthentication
 
+from mreg.models import ExpiringToken
+
 EXPIRE_HOURS = getattr(settings, 'REST_FRAMEWORK_TOKEN_EXPIRE_HOURS', 8)
 
 
 class ExpiringTokenAuthentication(TokenAuthentication):
     def authenticate_credentials(self, key):
         try:
-            token = self.get_model().objects.get(key=key)
+            token = ExpiringToken.objects.get(key=key)
         except ObjectDoesNotExist:
             raise exceptions.AuthenticationFailed('Invalid token')
 
         if not token.user.is_active:
             raise exceptions.AuthenticationFailed('User inactive or deleted')
 
-        if token.created < timezone.now() - timedelta(hours=EXPIRE_HOURS):
+        if token.updated < timezone.now() - timedelta(hours=EXPIRE_HOURS):
             raise exceptions.AuthenticationFailed('Token has expired')
 
         # Update the created time of the token to keep it valid.
-        token.created = timezone.now()
+        token.updated = timezone.now()
         token.save()
 
         return token.user, token
diff --git a/mreg/models.py b/mreg/models.py
index 1888364..26babd9 100644
--- a/mreg/models.py
+++ b/mreg/models.py
@@ -10,6 +10,7 @@ from django.db import DatabaseError, models, transaction
 from django.db.models import Q
 from django.utils import timezone
 from netfields import CidrAddressField, NetManager
+from rest_framework.authtoken.models import Token
 
 from .fields import DnsNameField, LCICharField
 from .models_auth import User  # noqa: F401, needed by mreg.settings for now
@@ -831,3 +832,6 @@ class BACnetID(models.Model):
             else:
                 return j
         return j
+
+class ExpiringToken(Token):
+    updated = models.DateTimeField(auto_now_add=True)

```

...which I *believe* is the correct approach (improvements welcome!); however we also have to override all the methods of [`rest_framework.authentication.TokenAuthentication`](https://github.com/encode/django-rest-framework/blob/master/rest_framework/authentication.py#L151).

Can you think of an easier way to solve this problem?  At this level it's tempting to just load [`knox`](https://github.com/James1345/django-rest-knox/blob/master/knox/auth.py#L25) instead.

oyvindkolbu: I think you are almost there. From:
https://github.com/encode/django-rest-framework/blob/61e7a993bd0702d30e3049179000bc7c5f284781/rest_framework/authentication.py#L164-L168
it seems easy to just set `self.model = ExpiringToken`. Probably easiest via a `__init__` method.

oyvindkolbu: Also would be nice with some tests.

ponas: For some reason I expected `updated` to be part of the `Token` model already – but it's just a common thing to include in custom abstract base classes. Maybe use the opportunity to name it something else, like `last_used`, `last_seen` or similar.

oyvindkolbu: I also assumed `updated` was an existing field when you mentioned it, but `last_used` sounds OK.

mbakke: > I think you are almost there. From:
> 
> https://github.com/encode/django-rest-framework/blob/61e7a993bd0702d30e3049179000bc7c5f284781/rest_framework/authentication.py#L164-L168
> 
> it seems easy to just set `self.model = ExpiringToken`. Probably easiest via a `__init__` method.

Great catch, thank you!  That worked like a charm.  I've pushed a new commit that introduces `ExpiringToken`.

mreg/api/v1/tests/tests.py
```diff
@@ -193,12 +194,28 @@ def test_logout_without_authentication(self):
...
+                                  "password":"test"})
+        new = ExpiringToken.objects.get(user=self.user)
+        assert old.key != new.key
+
+    def test_token_usage(self):
+        old = ExpiringToken.objects.get(user=self.user)
+        self.assert_get("/hosts/")
+        new = ExpiringToken.objects.get(user=self.user)
+        assert old.created == new.created
+        assert old.last_used != new.last_used
```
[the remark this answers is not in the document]
mbakke: Fixed, thanks!

mreg/models.py
```diff
@@ -831,3 +832,6 @@ def first_unused_id() -> int:
             else:
                 return j
         return j
+
+class ExpiringToken(Token):
+    last_used = models.DateTimeField(auto_now_add=True)
```
oyvindkolbu, line 837: Should use `auto_now=True`.

mreg/authentication.py
```diff
@@ -20,7 +27,11 @@ def authenticate_credentials(self, key):
         if not token.user.is_active:
             raise exceptions.AuthenticationFailed('User inactive or deleted')
 
-        if token.created < timezone.now() - timedelta(hours=EXPIRE_HOURS):
+        if token.last_used < timezone.now() - timedelta(hours=EXPIRE_HOURS):
             raise exceptions.AuthenticationFailed('Token has expired')
 
+        # Update the created time of the token to keep it valid.
+        token.last_used = timezone.now()
```
oyvindkolbu, line 34: This is automatically done with a `auto_now=True` `DateTimeField`.
Keep to comment, but add that it doesn't need to explicitly update the field. This is similar to the `BaseModel`'s `updated_at` field.

oyvindkolbu (changes requested):

mbakke:
Errh, I did not try logging into an existing mreg installation after this change, and just blindly trusted the tests.

Logging in with a user that already has a token causes a crash:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "authtoken_token_user_id_key"                                                                                       
DETAIL:  Key (user_id)=(2) already exists.
```

Demoted to draft status until I find a solution.

oyvindkolbu:
> Errh, I did not try logging into an existing mreg installation after this change, and just blindly trusted the tests.

Possibly because it already contained a token from created by `Token` and not `TokenExpire`, therefore you need to 
add a similar cleanup of `Token` here
https://github.com/unioslo/mreg/pull/442/files#diff-98101bc9229886064733bd64cbf050fde15778a4361b28382277fa40f295d3eaR30

The tests don't have stale data in them, so thats why it didn't catch it.

Could easily be provoked by adding a `Token.object.create...` before the `post` in `test_token_rotation`. Also check that it is gone afterwards. Probably better as a dedicated test.

mreg/authentication.py
mbakke: Using `auto_now` makes a lot of sense.  I had to mock the current time in order to make the `test_force_expire` test pass because we no longer control the `last_used` field, and added a development dependency on `mock`.

mbakke:
> Could easily be provoked by adding a `Token.object.create...` before the `post` in `test_token_rotation`. Also check that it is gone afterwards. Probably better as a dedicated test.

Right, thanks for the hint.  I had to override `get_token_client` to prevent `ExpiringToken` from being issued first and made a new test class:

```
diff --git a/mreg/api/v1/tests/tests.py b/mreg/api/v1/tests/tests.py
index 9790561..4140ebe 100644
--- a/mreg/api/v1/tests/tests.py
+++ b/mreg/api/v1/tests/tests.py
@@ -238,6 +238,39 @@ class APITokenAuthenticationTestCase(MregAPITestCase):
         self.assert_post_and_400("/api/token-auth/", {"who":"someone","why":"because"})
         self.assert_post_and_400("/api/token-auth/", {})
 
+
+from rest_framework.authtoken.models import Token
+class MigratingTokenTestCase(MregAPITestCase):
+    """ Ensure that an existing Token is replaced by ExpiringToken. """
+    def get_token_client(self, username=None, superuser=True, adminuser=False):
+        if username is None:
+            if superuser:
+                username = 'superuser'
+            elif adminuser:
+                username = 'adminuser'
+            else:
+                username = 'nobody'
+        self.user = get_user_model().objects.create_user(username=username,
+                                                         password="test")
+        self.user.groups.clear()
+        token, created = Token.objects.get_or_create(user=self.user)
+        if superuser:
+            self.add_user_to_groups('SUPERUSER_GROUP')
+        if adminuser:
+            self.add_user_to_groups('ADMINUSER_GROUP')
+        client = APIClient()
+        client.credentials(HTTP_AUTHORIZATION='Token ' + token.key)
+        return client
+
+    def test_conflicting_token(self):
+        self.assert_post_and_401("/api/token-auth/",
+                                 {"username": self.user,
+                                  "password":"test"})
+        self.assertFalse(Token.objects.get(user=self.user))
+
 class APIAutoupdateZonesTestCase(MregAPITestCase):
     """This class tests the autoupdate of zones' updated_at whenever
        various models are added/deleted/renamed/changed etc."""
```

...however the request never makes it to our view in the test and is stopped somewhere in [`APIView`](https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L104) AFAICT. When running the server normally, clearing the `Token` inside `ObtainExpiringAuthToken` works.  Any idea what can cause this discrepancy?

mbakke:
To clarify, this does work in practice:

```
diff --git a/mreg/api/views.py b/mreg/api/views.py
index adacb03..9833e66 100644
--- a/mreg/api/views.py
+++ b/mreg/api/views.py
@@ -9,6 +9,8 @@ from rest_framework import serializers
 from rest_framework.exceptions import AuthenticationFailed
 
 from mreg.models import ExpiringToken
+# Include the "vanilla" DRF Token in order to migrate existing installations.
+from rest_framework.authtoken.models import Token
 
 
 class ObtainExpiringAuthToken(ObtainAuthToken):
@@ -26,6 +28,9 @@ class ObtainExpiringAuthToken(ObtainAuthToken):
 
         user = serializer.validated_data['user']
 
+        # Delete any "old-style" DRF Token to prevent a conflict.
+        Token.objects.filter(user=user).delete()
+
         # Force token rotation.
         ExpiringToken.objects.filter(user=user).delete()
 

```

...but I haven't been able to write a test for it.

I wonder if we can take a "shotgun" approach and just clear *all* tokens at server startup?

oyvindkolbu:
> To clarify, this does work in practice:
> 
> ```
> diff --git a/mreg/api/views.py b/mreg/api/views.py
> index adacb03..9833e66 100644
> --- a/mreg/api/views.py
> +++ b/mreg/api/views.py
> @@ -9,6 +9,8 @@ from rest_framework import serializers
>  from rest_framework.exceptions import AuthenticationFailed
>  
>  from mreg.models import ExpiringToken
> +# Include the "vanilla" DRF Token in order to migrate existing installations.
> +from rest_framework.authtoken.models import Token
>  
>  
>  class ObtainExpiringAuthToken(ObtainAuthToken):
> @@ -26,6 +28,9 @@ class ObtainExpiringAuthToken(ObtainAuthToken):
>  
>          user = serializer.validated_data['user']
>  
> +        # Delete any "old-style" DRF Token to prevent a conflict.
> +        Token.objects.filter(user=user).delete()
> +
>          # Force token rotation.
>          ExpiringToken.objects.filter(user=user).delete()
>  
> ```

This makes sense.

> I wonder if we can take a "shotgun" approach and just clear _all_ tokens at server startup?

A crude but efficient solution. All old tokens will be invalidated on first usage anyhow, so
should be fine.

mbakke:
> > I wonder if we can take a "shotgun" approach and just clear _all_ tokens at server startup?
> 
> A crude but efficient solution. All old tokens will be invalidated on first usage anyhow, so
> should be fine.

It turns out that doing database operations during initialization is unsafe and discouraged:

 https://docs.djangoproject.com/en/dev/ref/applications/#django.apps.AppConfig.ready

Instead I opted to add a management command to delete all tokens.  It should be run before starting the server, and only needs to be done once.  I also added it to the Docker image and will remove once its consumers are updated.

Oh, and there is a new Admin view to inspect the new `ExpiringToken`.

oyvindkolbu (approved):

PR #455, opened 2021-09-24
Speed up forward zones by fetching fewer Host fields.

oyvindkolbu:
As only name and ttl are needed, only fetch them and ignore the rest.
Also use the iterator function as it does not cache anything, and the
data is only used once.


mbakke (approved):
LGTM :+1:

PR #456, opened 2021-09-24
Avoid delete/create race by only deleting tokens which are expired

### LINKED ISSUE #454: Deadlocks when the same user operates concurrently
When multiple clients try to authenticate and work against mreg at the same time with the same user, mreg may reach an Internal Server Error due to a database deadlock when one client updates `mreg_expiringtoken.last_used` at the same time another login of the same user gets issued a new token.

```
"deadlock detected","Process 2180545 waits for ShareLock on transaction 1481202488; blocked by process 2180546.
Process 2180546 waits for ShareLock on transaction 1481202487; blocked by process 2180545.
Process 2180545: DELETE FROM ""authtoken_token"" WHERE ""authtoken_token"".""key"" IN ('XXXX')
Process 2180546: UPDATE ""mreg_expiringtoken"" SET ""last_used"" = '2021-09-24T09:04:01.419587+00:00'::timestamptz WHERE ""mreg_expiringtoken"".""token_ptr_id"" = 'XXXX'"
```

This is a regression from #442.

oyvindkolbu:
Check is the token is expired before deleting them. Better approach is to e.g. use django-rest-knox to give each
session an unique token.

Resolves #454

### DIFF
--- a/mreg/api/v1/tests/tests.py
+++ b/mreg/api/v1/tests/tests.py
@@ -204,13 +204,30 @@ class APITokenAuthenticationTestCase(MregAPITestCase):
             mock_future.return_value = minute_after_expiry
             self.assert_get_and_401("/hosts/")
 
-    def test_token_rotation(self):
+    def test_token_reissue(self):
+        """ Authenticating before expiry should issue the same token. """
         old = ExpiringToken.objects.get(user=self.user)
         self.assert_post_and_200("/api/token-auth/",
                                  {"username": self.user,
                                   "password":"test"})
         new = ExpiringToken.objects.get(user=self.user)
-        assert old.key != new.key
+        assert old.key == new.key
+        self.assertLess(old.last_used, new.last_used)
+
+    def test_token_rotation(self):
+        old = ExpiringToken.objects.get(user=self.user)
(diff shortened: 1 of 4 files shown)


mreg/api/v1/tests/tests.py
```diff
@@ -204,13 +204,30 @@ def test_force_expire(self):
...
+
+    def test_token_rotation(self):
+        old = ExpiringToken.objects.get(user=self.user)
+        EXPIRE_HOURS = getattr(settings, 'REST_FRAMEWORK_TOKEN_EXPIRE_HOURS', 8)
+        minute_after_expiry = old.last_used + timedelta(hours=EXPIRE_HOURS,
+                                                        minutes=1)
+        with mock.patch('django.utils.timezone.now') as mock_future:
+            mock_future.return_value = minute_after_expiry
+            # Remove old Authorization header, otherwise the expired Token will be rejected with a 401
+            self.client.credentials()
```
mbakke, line 225: Good catch, that would have taken me some time to figure out :)

mbakke (approved):
Adding an `is_expired` property is a great idea.  Thanks for fixing my hasty PR :+1:

PR #457, opened 2021-11-16
OpenAPI schema generation

mbakke:
These patches make `./manage.py generateschema` work, and inserts it + the resulting `openapi.yml` into the CI pipeline.

It also adds Python 3.10 to the test matrix.


.github/workflows/test.yml
```diff
@@ -57,10 +58,18 @@ jobs:
...
       - name: Test
         run: coverage run manage.py test -v2
         env:
           MREG_DB_PASSWORD: postgres
+      - name: Upload OpenAPI schema
+        if: matrix.python-version == '3.10'
+        uses: actions/upload-artifact@v2
+        with:
+          name: openapi.yml
+          path: openapi.yml
```
oyvindkolbu, line 72: Where does this end up and what is it used for?

.github/workflows/test.yml
```diff
@@ -80,6 +89,7 @@ jobs:
           - '3.7'
           - '3.8'
           - '3.9'
+          - '3.10'
```
oyvindkolbu, line 92: 👍

oyvindkolbu (approved):
:shipit:

.github/workflows/test.yml
mbakke: Currently it is only stored as a temporary artifact in GitHub Actions, see e.g. [here](https://github.com/mbakke/mreg/actions/runs/1468068325).  Mostly so external developers can access it without setting up a local development environment or scraping the live server.

This file can be used to generate a complete and pretty API overview (see [here](https://safeaim.github.io/mreg-openapi-jekyll/) for an example), or even for automated generation of API clients.  I.e. `mreg-cli` could be rewritten to generate types, methods, etc boilerplate from the OpenAPI spec.

That is to say, I'm not sure what to do with this particular artifact yet.  It could make sense to replace the `/docs` endpoint with a static site built on this, to avoid pulling in `uritemplate` and `pyyaml` at runtime, for example.  We'll see how the internal demand plays out.

PR #458, opened 2021-12-06
requirements: Turn uritemplate and pyyaml into runtime dependencies.

mbakke:
...so that OpenAPI schema generation works for users running in a venv.


oyvindkolbu (approved):
:shipit:

PR #460, opened 2022-01-24
Update dependencies

mbakke:
Some minor dependency updates, preparing for Django 4.0.

(mreg works fine with Django 4, but the current version of django-url-filter currrently does not)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,12 @@
-Django==3.2.5
-djangorestframework==3.12.4
-django-auth-ldap==2.4.0
+Django==3.2.11
+djangorestframework==3.13.1
+django-auth-ldap==4.0.0
 django-logging-json==1.15
-django-netfields==1.2.2
+django-netfields==1.2.4
 django-url-filter==0.3.15
 gunicorn==20.1.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,12 +1,12 @@
-Django==3.2.5
-djangorestframework==3.12.4
-django-auth-ldap==2.4.0
+Django==3.2.11
+djangorestframework==3.13.1
+django-auth-ldap==4.0.0
```
oyvindkolbu, line 3: Tested auth-ldap locally? That is the one big thing which is not currently automatically tested.

mbakke: I haven not (apart from running its test suite); but I read the changelog and don't expect any surprises:

*  https://github.com/django-auth-ldap/django-auth-ldap/blob/master/docs/changes.rst
*  https://github.com/django-auth-ldap/django-auth-ldap/compare/2.4.0...4.0.0

Adding integration tests for LDAP connectivity should not be Too Hard, methinks...  I'll open an issue for it.

oyvindkolbu (approved):
:shipit:

PR #469, opened 2022-05-21
CI: Run 'apt-get update' before installing dependencies.

mbakke:
The [Actions documentation](https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners) recommends running `apt-get update` before using `apt`.

This should solve the error in #468.

While at it, ignore CI on README changes.


oyvindkolbu (approved):
:shipit:

PR #473, opened 2022-11-02
Publish events to RabbitMQ

oyvindhagberg:
Events are sent to MQ when:
- Hosts are added or removed
- Roles or atoms are created or deleted
- Atoms are added to or removed from roles

### DIFF
--- a/mreg/mqsender.py
+++ b/mreg/mqsender.py
@@ -1,0 +1,72 @@
+from django.conf import settings
+
+from datetime import datetime, timezone;
+import json
+import os
+import pika
+import random
+import ssl
+import string
+import time
+
+from pika.exceptions import (ConnectionClosedByBroker, StreamLostError, AMQPConnectionError)
+
+
+class MQSender(object):
+
+	def __new__(cls):
+		if not hasattr(cls, 'instance'):
+			cls.instance = super(MQSender, cls).__new__(cls)
+		return cls.instance
+
+	def __init__(self):
+		self.mq_channel = None
+		self.mq_id : int = int(time.time_ns()/1000000)
+
+	def send_event(self, obj, routing_key):
+		config = getattr(settings, 'MQ_CONFIG', None)
+		if config is None:
+			return
+
+		# Add an id property to the event
+		obj['id'] = self.mq_id
+		self.mq_id += 1
+
+		# Add a timestamp to the event
+		local_time = datetime.now(timezone.utc).astimezone()
+		obj['timestamp'] = local_time.isoformat()
+
+		for retry in range(10):
+			if self.mq_channel is None or self.mq_channel.connection.is_closed:
+				credentials = pika.credentials.PlainCredentials(
+					username=config['username'],
+					password=config['password'],
+				)
+				ssl_options = None
+				if config.get('ssl',False):
+					ssl_context = ssl.create_default_context()
+					ssl_options = pika.SSLOptions(ssl_context, config['host'])
+				connection_parameters = pika.ConnectionParameters(
+					host=config['host'],
+					credentials=credentials,
+					ssl_options = ssl_options,
+					virtual_host = config.get('virtual_host','/'),
+				)
+				try:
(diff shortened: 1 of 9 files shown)


mreg/utils.py
```diff
@@ -130,3 +140,59 @@ def normalize_mac(mac: str) -> str:
     mac = re.sub('[.:-]', '', mac).lower()
     return ":".join(["%s" % (mac[i:i+2]) for i in range(0, 12, 2)])
 
+
+mq_channel = None
+mq_id : int = 1
+pid_and_time = str(os.getpid()) + " " + str(time.time()) + " "
+
+def send_event_to_mq(obj, routing_key):
```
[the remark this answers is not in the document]
oyvindkolbu: Helt enig med @mbakke om egen `mq.py`, samt gjerne ting egen klasse.

mreg/utils.py
```diff
@@ -130,3 +140,59 @@ def normalize_mac(mac: str) -> str:
     mac = re.sub('[.:-]', '', mac).lower()
     return ":".join(["%s" % (mac[i:i+2]) for i in range(0, 12, 2)])
 
+
+mq_channel = None
+mq_id : int = 1
+pid_and_time = str(os.getpid()) + " " + str(time.time()) + " "
```
oyvindkolbu, line 146: Hvorfor statisk variabel med et fast tidspunkt?

mreg/utils.py
```diff
@@ -130,3 +140,59 @@ def normalize_mac(mac: str) -> str:
...
+    config = getattr(settings, 'MQ_CONFIG', None)
+    if config is None:
+        return
+
+    global mq_channel
+    global mq_id
+    global pid_and_time
+
+    # Add an id property to the event
+    obj['id'] = hashlib.md5((pid_and_time + str(mq_id)).encode('utf-8')).hexdigest()
```
oyvindkolbu, line 158: Med mindre dette må være en md5-digest, så bør ikke noe mye kode bruke md5, samme hvor trivielt.

mreg/utils.py
oyvindhagberg: Så koden slipper å kalle opp os.getpid() og time.time() hver gang en event oppstår.
Og grunnen til at pid og tidspunktet er en del av event-id'ene er at hvis mreg restartes så vil vi unngå at event-id'er kan gjenta seg. Samtidig vil vi unngå å ha et løpenummer som vi må lagre noe sted når mreg avsluttes.

oyvindhagberg: Endret i 76759f2

mreg/utils.py
oyvindhagberg: Fjernet md5 i 76759f2

mreg/mqsender.py
```diff
@@ -0,0 +1,72 @@
...
+class MQSender(object):
+
+	def __new__(cls):
+		if not hasattr(cls, 'instance'):
+			cls.instance = super(MQSender, cls).__new__(cls)
+		return cls.instance
+
+	def __init__(self):
+		self.mq_channel = None
+		self.mq_id : int = int(time.time_ns()/1000000)
```
oyvindkolbu, line 24: No need to force the int conversion, use instead:
`time.time_ns()//1_000_000`

oyvindkolbu (approved):
Looks better now. Though tests would be nice, that would require a fair bit of mocking.

oyvindkolbu (approved):
:shipit:

ponas:
Some downsides to this implementation:
* if the message broker is having performance issues, it will latency to all model changes / API requests
  * best case: the connection is retried ten times
  * worst case: the connection hangs, waiting for a timeout
* if the message broker is down, messages are silently discarded

Suggestion:
* add an intermediate `Notification` model for serialized message data
* change the signals to create `Notification` objects
* add a long-running process (via a management command) that:
  * selects the N oldest notifications
  * publishes them to the message broker
  * deletes the notification if successful
  * sleeps, repeats

With this, an auto-incrementing primary key could be used as the message ID.

PR #481, opened 2023-03-14
Add a new workflow that builds with a Dockerfile

oyvindhagberg:
This PR adds a new workflow that does the following:
- Builds a container image using a Dockerfile
- Tests the image by starting a container and running `manage.py test` inside it
- Publishes the image to GitHub Packages

The badge in the `README.md` file is also updated to reflect the build status of the new workflow.
The workflow aims to replace the one found in `docker-image.yml`.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -1,12 +1,77 @@
 name: Container image
-
 on:
-  workflow_dispatch:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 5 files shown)


terjekv (approved):
Looks fine to me. Long(er) term there are a few things to think about, but these are issues for later.

1) Add support beyond x86_64 via buildx and friends.
2) Shipping a full environment via Docker compose / Helm.

PR #490, opened 2023-03-28
Fewer dependencies for the container image

oyvindhagberg:
This PR does the following:
- The entrypoint in the container image will create the CITEXT extension in the database if needed. Previously, that had to be done before starting the container.
- The container image is built with default mregsite and logs directories, so it can run without these being mounted into it. Previously it would exit unless these were mounted in.
- A docker-compose.yml file is added to the repo.
- The section about Docker in the README file is updated. Also add names of more contributors.
- ~The container-image.yml workflow is modified to trigger on pushes to all branches, but only push the built image to ghcr.io if the branch is master. This way, we get to test that workflow on new code before merging into master.~
*Done in https://github.com/unioslo/mreg/commit/749d0968b034ee982c91ee17d37df3d04e3262c7*

### DIFF
--- a/mreg/management/commands/create_citext_extension.py
+++ b/mreg/management/commands/create_citext_extension.py
@@ -1,17 +1,38 @@
 from django.core.management.base import BaseCommand, CommandError
 from django.db import connection
 from sys import stdout
+from django.conf import settings
+from psycopg2 import connect
 
 class Command(BaseCommand):
 	help = 'Create the CITEXT extension in the database.'
 
-	def handle(self, *args, **kwargs):
+	def add_arguments(self, parser):
+		# optional argument
+		parser.add_argument(
+			'--database',
+			type=str,
+			help='Database name',
+		)
+
+	def handle(self, *args, **options):
 		stdout.write("Attempting to create the CITEXT extension in the database...\n")
 		stdout.flush()
 		try:
-			with connection.cursor() as cursor:
+			con = connection
+			if options['database']:
+				stdout.write(f"Connecting to database {options['database']}\n")
+				stdout.flush()
+				con = connect(host=settings.DATABASES['default']['HOST'],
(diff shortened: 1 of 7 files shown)


terjekv:
This got me thinking... Should we name and/or tag the artifacts better? Naming them something like `release` and `master`, while tagging with release number and commit hash, so it's a bit clearer what container one is running?

oyvindhagberg:
> This got me thinking... Should we name and/or tag the artifacts better? Naming them something like `release` and `master`, while tagging with release number and commit hash, so it's a bit clearer what container one is running?

Good idea, let's do that, but not in this pull request. I created [an issue](https://github.com/unioslo/mreg/issues/493)

mreg/management/commands/create_citext_extension.py
```diff
@@ -0,0 +1,38 @@
+from django.core.management.base import BaseCommand, CommandError
+from django.db import connection
+from sys import stdout
+from django.conf import settings
+from psycopg2 import connect
+
+class Command(BaseCommand):
```
terjekv, line 7: This is brilliant, but we should point to some documentation with regards to the dependencies (rpm packages/libraries/etc) for installing CITEXT into a database.

terjekv (approved):
Looks good to me. The create_citext_extension is great to have.

mreg/management/commands/create_citext_extension.py
oyvindhagberg: Have a look at 121566ee99dc1804942d87f136fe9dab8307624a, that'll point users in the right direction.

terjekv: Right, don't mind me. :)

terjekv (approved):

terjekv (approved):

PR #491, opened 2023-03-29
Django 4 support: Option 5, migrating to django-filter

### LINKED ISSUE #487: Support for Django4.
The main show-stopper for Django 4 support is the use of [django-url-filter](https://pypi.org/project/django-url-filter/0.3.15/). The original code base isn't maintained: https://github.com/miki725/django-url-filter, but there are two solid forks, both available on PyPI:

### dj-url-filter

- https://github.com/enjoy2000/django-url-filter.
- [PyPi package](https://pypi.org/project/dj-url-filter/).
- Retains support for python2.7 (causing a dependency on [six](https://pypi.org/project/six/)).
- Retains support for SQLAlchemy, and other backends.
- Sigh, we might have six and 2.7 support, but the [package from PyPi](https://pypi.org/project/dj-url-filter/) is marked as requiring Python 3.9+.

### django-ufilter

- https://github.com/Qu4tro/django-ufilter.
- [PyPI package](https://pypi.org/project/django-ufilter/).
- Supports python3 only, lots of code cleanups.
- Code base looks good.
- Dedicated towards Django. 
- Migrated to poetry.
- The latest version, 0.4.0, which is the only version we'd want, is only installable with Django4+ for python versions 3.8+...
- https://django-ufilter.readthedocs.io/en/latest/history.html.

### This leaves us with the following options:

1. Continue to use django-url-filter and thus stay put at 3.2 indefinitely. Not ideal, 3.2 is EOL 01 Apr 2024: https://endoflife.date/django.
2. Migrate to django-ufilter and force us to python 3.8 and Django4 at the same time.
3. Migrate to dj-url-filter, allowing us to remain on Django3.2 for now, but forcing us to python3.9.
4. Migrate to dj-url-filter for now (again forcing us to python3.9), then look at migrating to django-ufilter when we drop support for Django3.
5. Migrate to [django-filter](https://django-filter.readthedocs.io/en/stable/), forcing us to manually write all the filters we have today.

terjekv:
Fixes #487 by implementing option 5 in the issue.

Supports: python3.7+, django 3.2+. Tests included for this range.
Breaking changes: None

This implementation uses django-filter as intended. It sets a default filter backend and uses standard filter_class models where possible.

Exceptions take place in the following views:
  - `HostList`. Relies on manipulating the queryset, and said manipulation is also used in the filter-less HostDetail.
  - `*Zone*List`. The abstraction model for these views is based around propagating a filterset into the parent class.

Concerns:
  - This patch creates explicit mapping filters for JSONField and CIDRField. It is unclear how well-tested these are.


terjekv:
Doing a close / reopen to see if tests will run.

terjekv:
Superseded by #492.

PR #496, opened 2023-05-11
Fix a bug in history.py / DjangoJSONModelEncoder

oyvindhagberg:
The bug was introduced in 014e25b6030a05adf9622694539b92a38f71a2b1 when removing what was thought to be unused code paths. Unfortunately, we didn't have a test that covered the particular case where this code path was used.
This commit adds a test and fixes the bug.


terjekv (approved):
Dang. We used it for policies. Go figure. Good catch.

PR #498, opened 2023-05-24
Create a local mreg superuser via a command.

### LINKED ISSUE #497: Create mreg superuser via a command.
To create a properly usable superuser, the solution right now is something like this:

````python
echo -ne '
from django.conf import settings
from django.contrib.auth import get_user_model
from django.contrib.auth.models import Group
user = get_user_model().objects.create_user(username="test",password="test")
user.groups.clear()
group, created = Group.objects.get_or_create(name="default-super-group")
group.user_set.add(user)
' | docker exec -i mreg python /app/manage.py shell
````

Django has a `python manage.py createsuperuser` that people may expect to work. We can either override createsuperuser or since we already have some mreg-related commands, create a superuser command of our own and add it to the mreg section.

### PR DESCRIPTION
**Pull request by terjekv**
Also see https://github.com/unioslo/mreg-cli/pull/165/files/2c3d49bfdba04c3a0f10bae57dd4545eee33c3a8#diff-c4e7046d5cc0897cad49c4f1b5da237873561d81cd3d0814638db9cf949fd24f for a typical use case.

Fixes #497

### DIFF
--- a/mreg/management/commands/create_mreg_superuser.py
+++ b/mreg/management/commands/create_mreg_superuser.py
@@ -1,0 +1,77 @@
+import getpass
+from sys import stdout
+
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
+from django.core.management.base import BaseCommand, CommandError
+
+
+class Command(BaseCommand):
+    """Create a local MREG user, by default with superuser privileges."""
+
+    help = "Create a local MREG user, by default with superuser privileges."
+
+    def add_arguments(self, parser):
+        """Add arguments to the command."""
+        parser.add_argument(
+            "--username",
+            type=str,
+            help="The username for the new user. If not provided, will be read from stdin.",
+        )
+        parser.add_argument(
+            "--password",
+            type=str,
+            help="The password for the new user. If not provided, will be read from stdin.",
+        )
+        parser.add_argument(
+            "--group",
+            type=str,
+            help=(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mreg/management/commands/create_mreg_superuser.py [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+        password = options["password"]
+        groupname = options["group"]
+
+        # The username and password will be a quoted str due to argparse type=str.
+        # This means that "not username" will work fine if the username is given as 0
+        # as argparse will deliver it as "0" and not int(0), which is false.
+        if not username:
+            username = input("Username: ")
+
+        if not username:
```
oyvindhagberg, line 54: I'm guessing this line should be `if not password:` :slightly_smiling_face:

mreg/management/commands/create_mreg_superuser.py
```diff
@@ -0,0 +1,77 @@
...
+            password = getpass.getpass("Password: ")
+
```
oyvindhagberg, line 56: Line 54 should probably be `if not password:`.
Other than that, looks good, I tested it and it works.

oyvindhagberg (changes requested):

mreg/management/commands/create_mreg_superuser.py
terjekv: Rebased and forced pushed a corrected version.

oyvindhagberg (approved):

PR #504, opened 2023-05-28
Send events to MQ when hosts are renamed.

### LINKED ISSUE #476: Send events to MQ when hosts are renamed

terjekv:
- Captures the old name with pre_save.
- Makes no assumtion on signal consistency in Django...

Fixes #476

### DIFF
--- a/mreg/signals.py
+++ b/mreg/signals.py
@@ -6,7 +6,13 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
 from django.db import transaction
-from django.db.models.signals import m2m_changed, post_delete, post_save, pre_delete, pre_save
+from django.db.models.signals import (
+    m2m_changed,
+    post_delete,
+    post_save,
+    pre_delete,
+    pre_save,
+)
 from django.dispatch import receiver
 
 from django_auth_ldap.backend import populate_user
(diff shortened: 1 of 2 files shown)


terjekv:
Clean up merge commit via rebase and force push.

mreg/signals.py
```diff
@@ -306,37 +358,79 @@ def update_hosts_when_zone_is_added(sender, instance, created, **kwargs):
...
+            obj = {
+                "old_host": old_name,
+                "new_host": instance.name,
+                "action": "host_updated",
+            }
+        else:
+            obj = {
+                "host": instance.name,
+                "action": "host_updated",
+            }
```
oyvindhagberg, line 426: - What do you think about using the same field name whether we have the old name or not? E.g. "host", and "old_host" in addition if it is not None. Might be less confusing to parse.
- How about using the event name "host_renamed" when the host has been renamed? Unless that may also happen without pre_save being called.

oyvindhagberg (changes requested):

mreg/signals.py
terjekv: 1. The reason I opted out of using the same field name (ie, `host`) is that I found it ambiguous. Is `host` the new hostname, or the old hostname the recipient knows? If you are going to update an entry based on the event, you need to use `old_host` if and only if `old_host` is set, as that's the hostname you know, but in every other version of the `host_updated` event, you would use `host` as the host you know. It's not a big deal, but I felt it being inconsistent.
2. We can certainly use a new event name. That would make using `host` above easier as we can define the API for the `host_rename` event.

oyvindhagberg: 1. That makes sense, thanks
2. Actually, I meant action and not event name. But on second thought let's keep the event name "host" and action "host_updated", that will allow us to extend the functionality in the future if there are other changes to hosts that we want to send events for.

oyvindhagberg (approved):

PR #511, opened 2023-06-16
Modify the container image workflow to run mreg-cli tests

oyvindhagberg:
This PR makes the following modifications to the "container image" workflow:
- Adds a job that runs the mreg-cli test suite towards the newly built mreg container image
- Modifies the titles of the workflow jobs to be more descriptive

Note that Python version 3.11 is hardcoded for mreg-cli here. There's no need to run this in a matrix, because mreg-cli has been tested with multiple python versions already, before making it to the "master" branch that we download it from.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   build:
-    name: Build
+    name: Build image
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
@@ -28,7 +28,7 @@ jobs:
         path: mreg.tgz
 
   test:
-    name: Test
+    name: Unit tests
     needs: build
     runs-on: ubuntu-latest
     services:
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image.yml
```diff
@@ -61,11 +61,38 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: mreg
+    - name: Load container image
+      run: |
+        docker load --input mreg.tgz
+        docker tag mreg ghcr.io/unioslo/mreg:latest  # trick the shell script
```
terjekv, line 76: I'm not sure what we're tricking here? :)

terjekv (approved):
Not sure about the tricking comment, but otherwise lit looks fine to me. Thanks!

.github/workflows/container-image.yml
```diff
@@ -61,11 +61,40 @@ jobs:
...
+    - name: Download artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: mreg
+    - name: Load container image
+      run: docker load --input mreg.tgz
+    - name: Tag container image
+      # There's a docker-compose.yml file in the mreg-cli repo that wants the image from ghcr.io,
+      # but we want to use the newly built custom image
+      run: docker tag mreg ghcr.io/unioslo/mreg:latest
```
oyvindhagberg, line 78: Cleaned it up a bit and wrote a better comment

PR #519, opened 2023-09-20
Fix curl auth issue

oyvindhagberg:
There is a strange error that occurs when trying to authenticate with `curl`, that doesn't happen when authenticating with `mreg-cli`. I tracked it down to the following code:
https://github.com/unioslo/mreg/blob/29005d08ebc66e4b22cac5bca8d3fc9aefe87917/mreg/log_processors.py#L42-L47

Apparently `event_dict["content"]` is sometimes a dict and sometimes a str.

This PR does the following:
- [Adds a step to the workflow to test auth with curl](https://github.com/unioslo/mreg/commit/a19b5d0ca249a8fd5ecc89b066abc04409d34f39)
- [Adds a type check to the filter_sensitive_data function to check whether the content is a dict or a str before attempting to replace the password with dots](https://github.com/unioslo/mreg/commit/e3fb5353b41c89cd5e1381b112319352a2804d82)

### DIFF
--- a/mreg/log_processors.py
+++ b/mreg/log_processors.py
@@ -2,6 +2,7 @@
 
 from collections import defaultdict
 from typing import Any
+import re
 
 from rich.console import Console
 from rich.text import Text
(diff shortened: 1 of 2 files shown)


mreg/log_processors.py
```diff
@@ -44,7 +45,10 @@ def filter_sensitive_data(_: Any, __: Any, event_dict: EventDict) -> EventDict:
         event: str = event_dict["event"]
 
         if event == "request" and "password" in content:
-            event_dict["content"]["password"] = '...'
+            if isinstance(event_dict["content"],dict):
+                event_dict["content"]["password"] = '...'
+            if isinstance(event_dict["content"],str):
```
terjekv, line 50: Should this be `elif`? It won't matter for the result, but it looks a bit weird to me that we run the test twice when only one result can happen (unless we in the previous test assign event_dict["content"] to a str.

terjekv (approved):
Looks good as a fix, but I am stumped as to why this happens.

mreg/log_processors.py
oyvindhagberg: Yes, it should. Fixed in https://github.com/unioslo/mreg/commit/a7c378b34c56933886db1bf946e6be50a73dc38e

PR #527, opened 2024-04-22
Fix build error by bumping base image versions in the Dockerfile

oyvindhagberg:
Recently, building the container images started to fail: https://github.com/unioslo/mreg/actions/runs/8784288442/job/24102213934

It's hard to know exactly when, because GitHub had disabled the Container Image workflow due to low activity in the repo. Regardless, the issue can be fixed by using newer versions of the base images.
I've done no further investigation (It works, so it's probably ok?)


pederhan (approved):
LGTM

PR #531, opened 2024-05-06
Modify container workflow to only publish on tags

oyvindhagberg:
This PR modifies the workflow that builds the container image so it will only be published if it is built from a tag, which will be the new way of defining a "production" version.
It also removes scheduled builds - I don't think we need them, since the image is really only used for CI and testing.
Finally, the PR updates github action versions.


terjekv (approved):
Looks like a standard way to trigger on tags. :)

PR #538, opened 2024-05-27
Handle JSON in Zone POST requests

pederhan:
Checks content type before trying to call `getlist` on request data.

### DIFF
--- a/mreg/api/v1/views_zones.py
+++ b/mreg/api/v1/views_zones.py
@@ -56,6 +56,13 @@ def _validate_nameservers(names):
         done.add(name)
 
 
+def _get_request_nameservers(request: Request, field: str = "primary_ns") -> list[str]:
+    """Extract nameservers from the request data."""
+    if request.content_type == "application/json":
+        return request.data.get(field, [])
+    return request.data.getlist(field, [])
+
+
 class ZoneList(generics.ListCreateAPIView):
     """
     get:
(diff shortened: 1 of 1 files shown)


oyvindhagberg:
I tested this with mreg-cli from the `migrate_to_pydandic` branch (248a19e) and verified that the "zone create" commands in the testsuite work now.

terjekv (approved):
```
- zone.primary_ns = request.data.getlist('primary_ns')[0]
+ zone.primary_ns = nameservers[0]
```

I'm not going to pretend that I understand why we're simply picking the first one, but the PR retains behaviour and fixes the issue with content type management. That works for me. :)

oyvindkolbu:
> ```
> - zone.primary_ns = request.data.getlist('primary_ns')[0]
> + zone.primary_ns = nameservers[0]
> ```
> 
> I'm not going to pretend that I understand why we're simply picking the first one, but the PR retains behaviour and fixes the issue with content type management. That works for me. :)

It is due to https://github.com/unioslo/mreg/blob/b4975799aaa18905e704182cead4f990aaad896a/mreg/models/zone.py#L58
We (ab)use the first name server in the list to be used as mname server in the SOA record for the zone.

The zone's `primary_ns` field could possibly be changed to `mname` to better reflect usage.

terjekv:
> We (ab)use the first name server in the list to be used as mname server in the SOA record for the zone.

Now that I know I wish I didn't know. :)

> The zone's `primary_ns` field could possibly be changed to `mname` to better reflect usage.

Yep.

PR #547, opened 2024-08-06
Further filter fixes.

terjekv:
Manually sets the appropriate filters for most filter sets. A number of attempts at automating the code led nowhere good, so this is where we are. 

A lot of this may be cleaned up when DRF 3.15.3 is released as we can bump DRF to 3.15 that officially supports Django5. We are stuck on 3.14.0 due to https://github.com/encode/django-rest-framework/issues/9358 but https://github.com/encode/django-rest-framework/pull/9483 has been merged an hopefully makes 3.15.3.

### DIFF
--- a/mreg/api/v1/tests/test_filtering.py
+++ b/mreg/api/v1/tests/test_filtering.py
@@ -23,6 +23,10 @@ class FilterTestCase(ParametrizedTestCase, MregAPITestCase):
             param("hosts", "name", "hostsname0.example.com", 1, id="hosts_name"),
             param("cnames", "host__name",  "cnameshostname1.example.com", 1, id="cnames_host__name"),
             param("cnames", "host__name__icontains",  "cnameshostnameicontains", 3, id="cnames_host__icontains"),
+            param("cnames", "host__name__iexact",  "cnameshostnameiexact1.example.com", 1, id="cnames_host__iexact"),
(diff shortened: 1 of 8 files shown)


mreg/api/v1/tests/test_filtering.py
```diff
@@ -0,0 +1,109 @@
...
+
+class FilterTestCase(ParametrizedTestCase, MregAPITestCase):
+    """Test filtering."""
+
+    # endpoint, query_key, target, expected_hits
+    #
+    # NOTE: The generated hostnames are UNIQUE across every test case!
+    # The format is: f"{endpoint}{query_key}{i}.example.com".replace("_", "")
+    # where i is the index of the hostname (and we make three for each test).
+    @parametrize(
```
pederhan, line 58: Should there be a test for `atoms__name__exact` here, or is that unnecessary?

terjekv: There should be a set of tests for policy. I'll add that.

terjekv: Fixed in https://github.com/unioslo/mreg/pull/547/commits/99fee8a2d5a58931d2cbf7bc4bf79b36dcd36b98.

pederhan (approved):
LGTM

PR #551, opened 2024-09-11
Allow control over IP allocation method during host create

terjekv:
Allow users to select either random or first free from a network when asking for an IP during host creation. Adds the optional field `allocation_method` to POST for `/hosts/` when asking for a network allocation, allowing the values of either "first" or "random". Defaults to first if not set.

The previous behaviour with `network` was random only.


mreg/types.py
```diff
@@ -0,0 +1,5 @@
+from enum import Enum
+
+class IPAllocationMethod(Enum):
```
pederhan, line 3: There is a case to be made for using `StrEnum` here, since we already build our container image for Python 3.11. `StrEnum` helps avoid bugs/visual noise from forgetting to use `.value` when displaying and comparing such Enums.

However, we still officially support 3.10, so yeah... This is fine for now.

terjekv: Yeah, I poked at StrEnum, but 3.10. :(

mreg/api/v1/tests/tests.py
```diff
@@ -557,6 +558,36 @@ def test_hosts_post_case_insenstive_201_created(self):
...
+        data['allocation_method'] = allocation_method
+        data['network'] = network
+        del data['ipaddress']
+        return data
+
+    def assert_successful_allocation(self, allocation_method):
+        data = self.prepare_allocator_data(allocation_method)
+        response = self.assert_post_and_201('/hosts/', data)
+        self.assertEqual(response['Location'], f'/api/v1/hosts/{self.post_data["name"]}')
+        # TODO: Implement validation for IP address allocation
```
pederhan, line 574: What is meant by this? That we ensure the IP that was allocated is valid IP in general? Is an IP in the given network? Doesn't overlap with an existing host's IP?

oyvindhagberg (approved):
LGTM :+1:

mreg/api/v1/tests/tests.py
terjekv: I'd love to be able to validate that we got the address we expected, but I don't really know how to do that with random allocation. :(

PR #513, opened 2023-06-30
Migrating away from postgres CI fields.

terjekv:
***Caveats:***

  - Assumes fields in LCI fields are all already in lowercase. No efforts are currently made to manipulate these fields. Enforcing lowercase during the migration is doable if required, but the existing implementation of these fields use `get_db_prep_save` to guarantee only lowercase data is written. 
  - Email fields are migrated to models.EmailField. Case insensitivity is not preserved.
  - Needs (much) more testing.

Implementation details:

  - Creates two LowerCase-fields to replace the LCI-fields, `LowerCaseCharField` and `LowerCaseDNSField`. The latter inherits from the former. These fields will, like the LCI fields before them, save their data in lowercase by using `get_db_prep_save`.
  - Uses a [manager](https://docs.djangoproject.com/en/4.2/topics/db/managers/) to hook into the calls to objects.get, objects.filter, and objects.exclude to ensure they all use lowercase searches into the LowerCase fields. This functionality relies on all fields that are to use lowercase inherit from mreg.fields.LowerCaseCharField.
  - Uses a mixin for views to overload get_object for relevant detail views. 
 
Noise:

  - Overloaded get_queryset() usage with prefetching has to manually use lower() on the relevant input due to the manager not applying to querysets. For django-filter, this is solvable by mixing in a filter class that does the lowercasing were applicable, but url_filter does not support this easily.

Fixes #489

PR #553, opened 2024-11-10
Use uv for dependency management and versioning

pederhan:
This PR adds dependency management backed by a [uv lockfile](https://docs.astral.sh/uv/concepts/projects/#project-lockfile). Building, installing and running the project has been fully migrated to [uv](https://docs.astral.sh/uv/) in CI as well.

## Development dependencies

Development dependencies are now specified as [PEP 735](https://peps.python.org/pep-0735/) dependency groups. Dependency groups are not yet supported by pip, so for now, this means the development environment can only be installed with uv. Mreg itself does not require uv.

## Docker image

Since uv [does not support](https://github.com/astral-sh/uv/issues/1681) the `pip wheel` interface, we still have to install pip in the build stage of the dockerfile to build wheels for all of the project's dependencies. This complicates the build stage somewhat. 

## Versioning

The project is now versioned with [`setuptools-scm`](https://pypi.org/project/setuptools-scm/), which uses git tags to determine the project version. This necessitates git being installed wherever the project is built (such as in the build stage of the dockerfile).

The project's version can be accessed via `mreg_cli.__about__.__version__` (or the short-hand import `mreg_cli.__version__`).

----

## Why not use `uv python install` in the Dockerfile?

https://docs.astral.sh/uv/guides/integration/docker/#installing-python-in-musl-based-images


entrypoint-test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 set -e
 cd /app
-./manage.py create_citext_extension --database template1
-./manage.py test --noinput --failfast
+uv run ./manage.py create_citext_extension --database template1
```
terjekv, line 4: Do we still need to create the citext extension? That should have been resolved in https://github.com/unioslo/mreg/pull/513

entrypoint.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/sh
 set -e
 cd /app
-./manage.py create_citext_extension
-./manage.py migrate
-#./manage.py runserver 0.0.0.0:8000
+uv run ./manage.py create_citext_extension
```
terjekv, line 4: Do we still need to create the citext extension? That should have been resolved in https://github.com/unioslo/mreg/pull/513

terjekv (approved):
LGTM.

entrypoint-test.sh
pederhan: Probably a regression from a bad rebasing job by me.

pederhan: Nevermind, that is still present in master. We can remove that here or in another PR.

Dockerfile
```diff
@@ -5,25 +5,33 @@ ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
 RUN apk update
-RUN apk add --virtual build-deps gcc python3-dev openldap-dev musl-dev
-RUN pip install --upgrade pip
-COPY requirements*.txt ./
-RUN pip wheel --no-cache-dir --wheel-dir /usr/src/mreg/wheels -r requirements.txt
+RUN apk add --virtual build-deps gcc python3-dev openldap-dev musl-dev git
+COPY . .
```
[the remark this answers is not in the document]
terjekv: I have done similar and have nothing remotely elegant to offer. I would suggest adding a comment as to both what and why as `COPY . .` looks absolutely bonkers.

pyproject.toml
```diff
@@ -1,25 +1,57 @@
...
+    "django-netfields==1.3.2",
+    "django-filter==24.2",
+    "structlog==24.1.0",
+    "rich==13.7.1",
+    "gunicorn==22.0.0",
+    "idna==3.7",
+    "psycopg2-binary==2.9.9",
+    "pika==1.3.2",
+    "sentry-sdk==2.3.1",
+    "tzdata==2024.1",
```
pederhan, line 22: Ideally we don't lock dependency versions here. That is what the lock file is for. These should be more permissive, so that we don't run into a dependency resolution problem.

I.e. multiple dependencies rely on `idna`, but in upgrading one of them, it bumps its own `idna` requirement to  `idna>=3.8`. That would create a dependency resolution conflict, since we require `idna==3.7` specifically.

The solution is to change all these from `==` to `>=`.

terjekv: I agree. Do we trust people to understand semantic versioning enough to just do major versions, or do we do major.minor? I'd ideally like to see major only, but I feel very confident in that something will break (see DRF).

pederhan: I think it's fine either way as long as we always use the lock file to install dependencies, and we are deliberate with our dependency upgrades.

We can upgrade each dependency individually with `uv sync --upgrade-package <package>`.

If we _know_, for example, Django 6 is broken, we can upgrade with `uv sync --upgrade-package "Django>=5, <6"` to ensure we don't upgrade to an incompatible version. 

There are probably some nuances here I'm missing, but the point is that _we_ control when we want to actually upgrade a package. The more relaxed version specifiers just avoids dependency resolution conflicts when we want to do so.

terjekv: Sounds good. `>=` it is!

pederhan:
`django-rest-framework` is still locked to 3.14.0, which has its own set of issues, but we will not resolve that in this PR. The container image unit tests will remain broken until we can fix that in a different PR.

PR #554, opened 2024-11-11
Support Django 5.1 and DRF 3.15.2

terjekv:
Larger rewrite to avoid issues with how serialises were used to create objects within transactions. Also requires a major rewrite of permissions... This is an exploratory PR.

- Signals or something is broken, zone generation fails.
- Probably other issues as well.


terjekv:
Replaced by https://github.com/unioslo/mreg/pull/602

PR #559, opened 2024-11-20
Fix tox environments

pederhan:
Fixes the `coverage` and `lint` tox environments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,6 +34,7 @@ basepython =
     python310: python3.10
     python311: python3.11
     python312: python3.12
+    python3.11
 commands =
     python --version
     python -m django version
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -34,6 +34,7 @@ basepython =
     python310: python3.10
     python311: python3.11
     python312: python3.12
+    python3.11
```
terjekv, line 37: Only 3.11? This seems very weird.

terjekv (approved):

tox.ini
pederhan: It's the fallback version for when no other version is found for the environment. Uv downloads and installs it. Do you want to run linting and coverage with 3 different interpreters each time?

How did it behave prior to the uv PR?

terjekv: I *think* coverage ran within the current env, not something by tox?

pederhan: I have changed it to just `python3` now, which should select any Python 3 dist (system or uv-managed) that is supported by the project.

PR #562, opened 2024-11-21
Meta endpoint expansion

terjekv:
BREAKING CHANGE:

`api/meta/versions` is now gone! (But this was never taken into use)

This has been replaced with `api/meta/version` (server version, open to authenticated users) and `api/meta/libraries` (core libraries, available to superusers and administrators).


terjekv:
This could do with some testing indeed. I'll poke at it.

terjekv:
Tests added, coverage OK. I'll do another coverage pass PR soon[tm].

pederhan (approved):
LGTM!

PR #564, opened 2024-11-21
Add support for filtering on ID for hostpolicy objects.

terjekv:
Fixes the missing support for filtering on ID for Roles and Atoms in HostPolicy.


pederhan (approved):
LGTM. I love Django filtering! 🤠

PR #565, opened 2024-11-25
Resolve issue with UserInfo crashing on permissions.

terjekv:
- JSON-ifying network addresses caused an exception. Casting it to str works.
 - Also fixes permission logic for the UserInfo endpoint.
 - Added testing of said endpoint with permissions.


pederhan (approved):
LGTM

PR #566, opened 2024-11-26
Refactoring tests for user permissions.

terjekv:


mreg/models/auth.py
```diff
@@ -10,3 +38,53 @@ def group_list(self):
...
+        return self.is_mreg_admin or self.is_mreg_superuser
+
+    @property
+    def is_mreg_superuser_or_hostpolicy_admin(self):
+        return self.is_mreg_hostpolicy_admin or self.is_mreg_superuser
+
+    @classmethod
+    def from_request(cls, request: HttpRequest) -> 'User':
+        if not request.user.is_authenticated:
+            raise exceptions.NotAuthenticated("Attempted to coerce an authenticated user from unauthenticated request")
```
pederhan, line 81: In order to not leak internal implementation details, could we instead log that message and raise a more generic 401 message?

terjekv: Yep. Done.

mreg/api/v1/tests/test_host_permissions.py
```diff
@@ -1,21 +1,41 @@
...
+    @skip("Overriding settings is broken")
+    def test_groups_always_a_list_from_str(self):
+        """Ensure that the settings are always a list, even if there is only one group."""
+        self.settings(SUPERUSER_GROUP='superuser')
+        self.assertEqual(MregAdminGroup.SUPERUSER.settings_groups_or_raise(), ['superuser'])
+
+    @skip("Overriding settings is broken")
+    def test_groups_always_a_list_from_list(self):
+        """Ensure that the settings are always a list, even if there is only one group."""
+        self.settings(SUPERUSER_GROUP=['superuser'])
```
ponas, line 37: This should work if used as a context provider, i.e.
```
        with self.settings(SUPERUSER_GROUP=['superuser']):
            self.assertEqual(MregAdminGroup.SUPERUSER.settings_groups_or_raise(), ['superuser'])
```
The alternative is to decorate the class or methods:
```
@override_settings(SUPERUSER_GROUP=['superuser'])
...
```

terjekv: Yay, thanks! The decorators worked fine!

pederhan (approved):
LGTM!

PR #568, opened 2024-12-04
Further coverage fixes, now with mocks.

terjekv:
Also reintroduces support for `# pragma: no cover`.


terjekv:
I have considered mocking more of the middleware, but it quickly gets to the point where I am testing the (magic)mock and not the middleware as there is so much that ends up being mocked.

oyvindhagberg (approved):
LGTM 👍

PR #569, opened 2024-12-05
Change how mreg-cli is used for testing in the workflow

oyvindhagberg:
This PR does the following:

- Modifies the container-image-workflow so it uses the `run_testsuite_and_record_V2.sh` script from mreg-cli
   - Upon a successful run of the tests, the workflow posts a comment on the commit about which version of mreg-cli was used for testing. This makes it easy to re-run these tests elsewhere.
- Adds a helper script that can be used with `git bisect run`
- Fixes a typo in Dockerfile
- Deletes old workflow- and guix-files that aren't in use anymore


.github/workflows/docker-image.yml
```diff
@@ -1,76 +0,0 @@
...
-          -S /app=app -S /etc/profile=etc/profile
-          --entry-point=bin/mreg-wrapper
-          -m ./ci/manifest.scm
-      - name: Upload artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: mreg-docker.tar.gz
-          path: mreg-docker.tar.gz
-
-  publish:
```
terjekv, line 51: Did we lose the entire publish step?

oyvindhagberg: Had a quick look at GitHub on my phone and it looks like the code isn't removed? Wasn't supposed to be removed anyway.. I'll have a look when I can

oyvindhagberg: Eh, the `docker-image.yml` workflow has been disabled for a long time now. It was superseded by `container-image.yml` and publishing is done there.

pederhan (approved):
LGTM!

PR #571, opened 2025-01-14
Add `meta/health/ldap` endpoint

pederhan:
This PR adds an endpoint for checking the status of the application's LDAP connection. It performs the following actions:

1. Tries to connect to LDAP server and bind using credentials defined in settings.
2. Responds with either 200 or 503 based on success of bind operation.

The response contains no data, just a status code. LDAP internals are not exposed.

## Tasks

- [x] Add `/meta/health/ldap` endpoint
- [x] Move heartbeat endpoint to `/meta/health/heartbeat`
- [x] Add tests

### DIFF
--- a/mreg/api/views.py
+++ b/mreg/api/views.py
@@ -219,81 +219,79 @@ class LDAPDetails:
         d["healthy"] = self.healthy
         return d
 
-class HealthLDAP(APIView):
 
+class HealthLDAP(APIView):
     permission_classes = (IsAuthenticated,)
 
     def get(self, request: Request):
         details = self._check_ldap_connection()
         st = status.HTTP_200_OK if details.healthy else status.HTTP_503_SERVICE_UNAVAILABLE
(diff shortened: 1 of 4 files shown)


mreg/api/views.py
```diff
@@ -199,3 +202,98 @@ def get(self, request: Request):
...
+            error_msg = f"LDAP connection error: {str(e)}"
+            logger.error(error_msg)
+            details.error = error_msg
+            return details
+
+        except Exception as e:
+            error_msg = f"Unexpected error during LDAP check: {str(e)}"
+            logger.error(error_msg)
+            details.error = error_msg
+            return details
```
pederhan, line 299: Need to use structured logging here

terjekv (changes requested):
We need a ChangeLog entry and we need to make it clear that the health endpoints are now public. :)

terjekv (approved):
418 - I'm a teapot. LGTM.

PR #572, opened 2025-01-22
Network policies and communities.

terjekv:
# Add support for networks to have policies and communities.

This PR allows for networks to have policies applied and communities attached to them. Network policies consists of attributes that say something useful to the end user about the network, while communities are groups of hosts within the network. 

These groups are typically used within networks with client isolation to talk east-west as well as north-south.

## Purpose

Say you want to create an isolated network, but within this network you want a set of communities that can talk together. You can do this creating a network attribute (`isolated`), and then create a network policy that includes that attribute. You then apply that policy to the network in question. Listing the network will now (hopefully) show the attribute.

Now, within the network itself, you may create two communities, `foo` and `bar`. You can then add hosts to these communities. Note that to add a host to a community, at least one of the IPs of the host has to be in the network that the policy the community belongs to.

## Changes

When creating a host, one may add a `network_community` attribute to the JSON, which takes the parameter of a network community ID that the host is to be assigned to. Note that this requires the host to be created with a network or an IP as a host cannot be a member of a network community that belongs to a network policy that again belongs to a network to which the host has no IPs.

## Removing / changing policies

Since hosts may belong to communities that belong to networks, trying to change IPs on a host now requires a check to see if the new IP can belong to the current community of the host (in effect, it has to be on the same network as before). If that isn't the case, we return a 409.


mreg/api/v1/views_network_policy.py
```diff
@@ -0,0 +1,171 @@
...
+    CommunitySerializer,
+    HostSerializer,
+)
+
+from mreg.api.v1.filters import NetworkPolicyAttributeFilterSet, NetworkPolicyFilterSet, CommunityFilterSet, HostFilterSet
+
+from mreg.api.v1.views import JSONContentTypeMixin
+from mreg.api.permissions import IsGrantedNetGroupRegexPermission, IsSuperOrNetworkAdminMember
+
+BASE_PATH = '/api/v1/networkpolicies/'
```
pederhan, line 19: Unused

mreg/models/host.py
```diff
@@ -21,6 +31,29 @@ class Meta:
...
+    def set_community(self, community: Community) -> bool:
+        """Set the community for this host.
+        
+        :param community: The community to set.
+        :return: True if the community was set, False otherwise
+        """
+        # We need to check that the community is applicable to the same
+        # network as one of the IP addresses of the host.
+        for ipaddress in self.ipaddresses.all(): # type: ignore
+            from mreg.models.network import Network
```
pederhan, line 46: Can probably move this import to the top of the method.

mreg/apps.py
```diff
@@ -1,8 +1,22 @@
...
+from django.conf import settings
+from django.db.models.signals import post_migrate
 
 
 class MregAppConfig(AppConfig):
     name = 'mreg'
 
     def ready(self):
         import mreg.signals # noqa
+        from .models.network_policy import NetworkPolicyAttribute
```
pederhan, line 11: Could be an absolute import like most other imports.

```suggestion
        from mreg.models.network_policy import NetworkPolicyAttribute
```

mreg/api/v1/views_network_policy.py
```diff
@@ -0,0 +1,249 @@
...
+
+        try:
+            network = Network.objects.get(network=network)
+        except Network.DoesNotExist:  # pragma: no cover
+            raise exceptions.NotFound("Network not found.")
+
+        # We do not have to worry about case sensitivity here, as the LowerCaseManager for the model will handle that.
+        try:
+            Community.objects.get(name=name, network=network)
+            raise ValidationError409(detail=f"Community with the name '{name}' already exists.")
```
pederhan, line 143: `raise` could be moved into `else` block to prevent subtle bugs stemming from raised error being caught by `except`. To be clear, it doesn't happen here, but should any of the exception definitions involved change, all bets are off.

mreg/models/host.py
terjekv: Done!

mreg/apps.py
terjekv: Yep. Done.

mreg/api/v1/views_network_policy.py
terjekv: Thanks, refactored.

mreg/apps.py
```diff
@@ -1,8 +1,22 @@
...
         import mreg.signals # noqa
+        from mreg.models.network_policy import NetworkPolicyAttribute
+
+        def create_protected_attributes(sender, **kwargs):
+            protected_attrs = getattr(settings, 'MREG_PROTECTED_POLICY_ATTRIBUTES', [])
+            for attr in protected_attrs:
+                NetworkPolicyAttribute.objects.get_or_create(
+                    name=attr,
+                    defaults={'description': 'Automatically created protected attribute.'}
+                )
```
pederhan, line 19: In the absence of any centralized config validation, we should check if we have a valid configuration here before creating the protected attributes. This prevents misconfiguration of the application such as:


```py
MREG_PROTECTED_POLICY_ATTRIBUTES = "isolated
```

Creating the attributes `i`, `s`, `o`, `l`, `a`, `t`, `e`, `d`.

```suggestion
        def create_protected_attributes(sender, **kwargs):
            protected_attrs = getattr(settings, 'MREG_PROTECTED_POLICY_ATTRIBUTES', [])

            if not isinstance(protected_attrs, list):
                raise ValueError('Config option MREG_PROTECTED_POLICY_ATTRIBUTES must be a list.')
            
            for attr in protected_attrs:
                NetworkPolicyAttribute.objects.get_or_create(
                    name=str(attr),
                    defaults={'description': 'Automatically created protected attribute.'}
                )
```

terjekv: Fixed universally.

mreg/api/v1/serializers.py
```diff
@@ -27,6 +30,49 @@ def validate(self, data):
...
 
 
+class CommunitySerializer(serializers.ModelSerializer):
+
+    # Members are all the hosts that have this community assigned.
+    hosts = serializers.SerializerMethodField(read_only=True)
+    global_name = serializers.SerializerMethodField(read_only=True)
+
+    def get_hosts(self, obj):
+        return [host.name for host in obj.hosts.all()]
```
oyvindkolbu, line 40: No need to create the objects when just seeking a value:
 ```return list(obj.hosts.values_list("name", flat=True))```

terjekv: Thanks! Fixed in https://github.com/unioslo/mreg/pull/572/commits/49a6907c8c16202cea473bd0d383aa930b781399

mreg/api/v1/endpoints.py
```diff
@@ -0,0 +1,13 @@
...
+class URL:
+    class NetworkPolicy(str, Enum):
+        LIST = "networkpolicy-list"
+        DETAIL = "networkpolicy-detail"
+        COMMUNITIES_LIST = "networkpolicy-communities-list"
+        COMMUNITY_DETAIL = "networkpolicy-community-detail"
+        COMMUNITY_HOSTS_LIST = "networkpolicy-community-hosts-list"
+        COMMUNITY_HOST_DETAIL = "networkpolicy-community-host-detail"
+        ATTRIBUTE_LIST = "networkpolicyattribute-list"
+        ATTRIBUTE_DETAIL = "networkpolicyattribute-detail"
```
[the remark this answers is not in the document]
terjekv: Yep. Fixed in https://github.com/unioslo/mreg/pull/572/commits/b965520c026f28533c6e85161c3493e3d1f9d70f.

pederhan (approved):
LGTM 🤠

PR #576, opened 2025-03-18
Expand default host output.

terjekv:
Most calls to the host endpoint is followed up by calls to a number of related APIs. This commit bundles a number of those calls into a single call to the host endpoint, drastically reducing the number of calls needed to create a host.

Requires https://github.com/unioslo/mreg/pull/572 due to serialiser additions.


pederhan (approved):
LGTM!

Issue #577, opened 2025-04-22
Proposal: Standardized error response schema for API


pederhan:
## The problem

Out of the box, DRF does not produce standardized errors. Depending on the source of the error and its type, the framework produces errors that do not follow a particular spec. The [documentation](https://www.django-rest-framework.org/api-guide/exceptions/#exceptions) provides examples of some of the errors it _may_ produce, but they are merely a guideline for what errors _can_ be, and not a definitive specification. There are no restrictions in place for how errors are structured.

This is obviously not great from a consumer perspective, since we have to know what kind of error we are dealing with before we can even begin to unravel what went wrong. 

## Solution: Use [drf-standardized-errors](https://github.com/ghazi-git/drf-standardized-errors)

When a standardized approach to error handling on the consumer side is needed, the official [DRF documentation](https://www.django-rest-framework.org/api-guide/exceptions/#third-party-packages) recommends using the package `drf-standardized-errors`. It produces errors that follow a single common spec:

```json
{
  "type": "validation_error",
  "errors": [
    {
      "code": "required",
      "detail": "This field is required.",
      "attr": "name"
    },
    {
      "code": "max_length",
      "detail": "Ensure this value has at most 100 characters.",
      "attr": "title"
    }
  ]
}
```

The obvious benefit here on the consumer side is the ability to define simple Pydantic models for parsing _all_ types of errors. A consumer could do something like this:

```py
from pydantic import BaseModel

class MREGError(BaseModel):
    code: str
    detail: str
    attr: str | None

class MREGHTTPError(BaseModel):
    type: str
    errors: list[MREGError] = []
```

Those few lines of code are enough to parse every kind of error the application produces, and it makes it easy to differentiate error types by just checking the top-level `"type"` field.

## Breaking changes

As outlined in https://github.com/unioslo/mreg-cli/pull/384#issuecomment-2820390750, using `drf-standardized-errors` would break error messages on failed login due to invalid username/password combination in mreg-cli, as well as any 3rd party scripts that rely on the JSON body of the current error responses MREG produces. 

In my opinion, those things are OK to break, as we never formally documented API error responses in the first place (I think?).

PR #578, opened 2025-04-22
Implement standardized API error response format

pederhan:
This PR adds [`drf-standardized-errors`](https://github.com/ghazi-git/drf-standardized-errors) as the default DRF exception handler, as outlined in #577.

## Examples

Adding an SRV record (`/api/v1/srvs/`) with an invalid name (`{"name": "whatever"}`):

```
{
  "type": "validation_error",
  "errors": [
    {
      "code": "invalid",
      "detail": "Must match: ^_[a-z0-9]+(([a-z0-9][_-]?)+[a-z0-9]+)?._(tcp|tls|udp)\\.",
      "attr": "name"
    }
  ]
}
```

Invalid username/password when logging in (`/api/token-auth`):

```
{
  "type": "client_error",
  "errors": [
    {
      "code": "authentication_failed",
      "detail": "Incorrect authentication credentials.",
      "attr": null
    }
  ]
}
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apk add --virtual build-deps gcc python3-dev openldap-dev musl-dev git
 # and we need to keep the .git directory intact.
 # The workaround is to copy everything, but limit it with .dockerignore.
 COPY . .
-COPY --from=ghcr.io/astral-sh/uv:0.4.30 /uv /uvx /bin/
+COPY --from=ghcr.io/astral-sh/uv:0.6.16 /uv /uvx /bin/
 RUN  uv venv \
     && uv sync --frozen --no-dev  \
     && uv export --no-hashes -o requirements.txt \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apk add --virtual build-deps gcc python3-dev openldap-dev musl-dev git
 # and we need to keep the .git directory intact.
 # The workaround is to copy everything, but limit it with .dockerignore.
 COPY . .
-COPY --from=ghcr.io/astral-sh/uv:0.4.30 /uv /uvx /bin/
+COPY --from=ghcr.io/astral-sh/uv:0.6.16 /uv /uvx /bin/
```
pederhan, line 16: Adding a new dependency (`uv add drf-standardized-errors`)  using a recent version of uv caused the revision of the lockfile to be updated to a version incompatible with uv 0.4.30. We may want to look into ways of ensuring uv version in the dev environment stays consistent with the one in the container image. For now, the versions are manually specified here and in `pyproject.toml`.

terjekv: ... For every problem with python there is a tool. This tool converts the original problem into another problem. Luckily, for this problem, there is a tool...

mreg/api/v1/tests/tests.py
```diff
@@ -869,7 +869,19 @@ def test_list(self):
 
     def test_post_must_have_both_fields_400_bad_request(self):
         ret = self.assert_post_and_400('/hinfos/', {'host': self.host.id, 'cpu': 'cpuname'})
-        self.assertEqual(ret.json(), {'os': ['This field is required.']})
+        self.assertEqual(
```
terjekv, line 872: Is... this the only place we test the format of errors? Charming! 

I'd like buy a type please.

terjekv (approved):
Great improvement tbh. The only thing is that we tie ourselves deeper into DRF...

PR #580, opened 2025-05-13
Add basic type validation/coercion of env vars in settings.py

pederhan:
For #572 

This PR adds rudimentary type validation and coercion of environment variable values. By specifying a default value, the value retrieved from the environment is coerced to the same type as the default value. If the value cannot be coerced to that type, the default value is returned instead.

This PR also exposes `MREG_MAP_GLOBAL_COMMUNITY_NAMES` as a boolean value (`1`, `true`, `y`, `yes`) via an environment value with the same name.


terjekv (approved):
Yay! LGTM!

PR #584, opened 2025-05-23
Add network admin check when creating hosts with network/broadcast addr

pederhan:
## Issue

Network administrators want to be able to create hosts with ip addresses that overlap with network or broadcast addresses. It has technically always been possible through the API, but we prevented users from doing so in the CLI. We want to allow network admins to do so through the CLI, and thus we need to implement this check server-side.

## Solution

Adds a permissions check when creating hosts with an ip address that is identical to the network's network or broadcast address. The user _must_ be an admin or a network admin in order to use such IP addresses when creating hosts.

### DIFF
--- a/mreg/api/permissions.py
+++ b/mreg/api/permissions.py
@@ -364,9 +364,8 @@ class IsGrantedReservedAddressPermission(IsAuthenticated):
         try:
             ipaddr = ipaddress.ip_address(ip)
         except ValueError:
-            raise exceptions.ValidationError(
-                {"ipaddress": "Invalid IP address."},
-            )
+            # invalid IP, let serializer handle it
+            return True 
 
         try:
             network: Network = Network.objects.get(network__net_contains=ip)
@@ -388,7 +387,7 @@ class IsGrantedReservedAddressPermission(IsAuthenticated):
     def has_destroy_permission(self, request: Request, view: ListCreateAPIView, validated_serializer: BaseModel):
         # Deleting will never assign IPs. 
         # Furthermore, the permissions check in `perform_destroy` passes 
-        # in an _instance_ instead of a serializer when checking destroy 
-        # permissions, thus we cannot access any sort of validated data.
+        # in a BaseModel _instance_ instead of a serializer when checking
+        # destroy permissions, so we cannot access any sort of validated data.
         return self.has_permission(request, view)
 
--- a/mreg/api/permissions.py
+++ b/mreg/api/permissions.py
@@ -17,7 +17,7 @@ from mreg.models.auth import User
 # the `DEFAULT_PERMISSION_CLASSES` we defined in `settings.py`, causing
 # an import cycle if we _actually_ import the generics module on runtime.
 if TYPE_CHECKING:
-    from rest_framework.generics import ListCreateAPIView
+    from rest_framework.generics import GenericAPIView
     from rest_framework.serializers import Serializer
     from mreg.models.base import BaseModel
 
@@ -334,7 +334,7 @@ class HostGroupPermission(IsAuthenticated):
 
 
 class IsGrantedReservedAddressPermission(IsAuthenticated):
-    def has_ipaddress_permission(self, request: Request, view: ListCreateAPIView, validated_serializer: Serializer):
+    def has_ipaddress_permission(self, request: Request, view: GenericAPIView, validated_serializer: Serializer):
         import mreg.api.v1.views
 
         user = User.from_request(request)
@@ -378,13 +378,13 @@ class IsGrantedReservedAddressPermission(IsAuthenticated):
                 )
         return True
 
-    def has_create_permission(self, request: Request, view: ListCreateAPIView, validated_serializer: Serializer):
+    def has_create_permission(self, request: Request, view: GenericAPIView, validated_serializer: Serializer):
         return self.has_ipaddress_permission(request, view, validated_serializer)
 
-    def has_update_permission(self, request: Request, view: ListCreateAPIView, validated_serializer: Serializer):
+    def has_update_permission(self, request: Request, view: GenericAPIView, validated_serializer: Serializer):
         return self.has_ipaddress_permission(request, view, validated_serializer)
 
-    def has_destroy_permission(self, request: Request, view: ListCreateAPIView, validated_serializer: BaseModel):
+    def has_destroy_permission(self, request: Request, view: GenericAPIView, validated_serializer: BaseModel):
         # Deleting will never assign IPs. 
         # Furthermore, the permissions check in `perform_destroy` passes 
         # in a BaseModel _instance_ instead of a serializer when checking
--- a/mreg/api/permissions.py
+++ b/mreg/api/permissions.py
@@ -335,23 +335,8 @@ class HostGroupPermission(IsAuthenticated):
 
 class IsGrantedReservedAddressPermission(IsAuthenticated):
     def has_ipaddress_permission(self, request: Request, view: GenericAPIView, validated_serializer: Serializer):
-        import mreg.api.v1.views
-
         user = User.from_request(request)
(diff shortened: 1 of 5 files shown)


mreg/api/v1/views.py
```diff
@@ -353,6 +355,36 @@ def get_queryset(self):
...
+
+        
+        try:
+            network: Network = Network.objects.get(network__net_contains=ip)
+        except Network.DoesNotExist:
+            raise ValidationError(
+                {"ipaddress": "No network found for the IP address."}
+            )
+        
+        if ipaddr in (network.network.broadcast_address, network.network.network_address):
```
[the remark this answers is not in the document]
pederhan: Done.

mreg/api/v1/views.py
```diff
@@ -353,6 +355,33 @@ def get_queryset(self):
...
+            ipaddr = ipaddress.ip_address(ip)
+        except ValueError:
+            raise ValidationError(
+                {"ipaddress": "Invalid IP address."},
+            )
+
+        try:
+            network: Network = Network.objects.get(network__net_contains=ip)
+        except Network.DoesNotExist:
+            pass # network not in mreg
```
terjekv, line 372: Wouldn't this then allow anyone to create any DNS record as long as the network is outside of mreg? I think this needs to be restricted to prevent someone from registering www.google.com. Ie, if the network isn't in mreg *or* we're asking for the presumptive broadcast *or* we're asking the network address, we need to be superuser_or_admin or network_admin (or probably is_mreg_superuser_or_hostpolicy_admin + is_mreg_network_admin)?

oyvindkolbu: No, this just skips the check for broadcast and network adresses because we don't know the network boundary.

Normal access is governed by `IsGrantedNetGroupRegexPermission`, so that should not be an issue with neither the allowed name nor IPs.

terjekv: Right, we're assuming that this code path is never hit without the view's permissions checks already having filtered who can deal with non-mreg networks. 

This does however mean that we're effectively performing permissions checking in two places, so we could consider moving the test for broadcast and network address assignments to the permissions code itself? I mostly find it iffy to have to look multiple places to check for permissions.

pederhan: You're right. I'll look into implementing it as a permissions class.

pederhan: Tests are passing with this new approach, but we need some new tests for this scenario. There are so many conditionals that make it return early with `True` that it's likely we don't even hit the relevant checks in the tests.

mreg/api/permissions.py
```diff
@@ -341,11 +341,17 @@ def has_ipaddress_permission(self, request: Request, view: ListCreateAPIView, va
...
+            view, 
+            (
+                mreg.api.v1.views.HostList,
+                mreg.api.v1.views.HostDetail,
+                mreg.api.v1.views.IpaddressDetail,
+                mreg.api.v1.views.IpaddressList,
+                mreg.api.v1.views.PtrOverrideDetail,
+                mreg.api.v1.views.PtrOverrideList
+            )
+        ):
```
pederhan, line 354: Can we remove this check? As long as we have an `"ipaddress"` that we mutate or assign somehow, we might want to check if the user is allowed to use such an IP? 

Then again, I'm sure doing that opens the door for some strange bugs down the road because we make too broad of an assumption about the semantics of such an `"ipaddress"` field.

mreg/api/permissions.py
```diff
@@ -364,9 +364,8 @@ def has_ipaddress_permission(self, request: Request, view: ListCreateAPIView, va
         try:
             ipaddr = ipaddress.ip_address(ip)
         except ValueError:
-            raise exceptions.ValidationError(
-                {"ipaddress": "Invalid IP address."},
-            )
+            # invalid IP, let serializer handle it
+            return True
```
pederhan, line 368: This might be a bit contentious...?

I think it makes more sense to pass the request on to other permission classes, and eventually the serializer, as opposed to throwing a 403 here.

mreg/api/permissions.py
```diff
@@ -317,3 +358,48 @@ def has_destroy_permission(self, request, view, validated_serializer):
...
+
+
+class IsGrantedReservedAddressPermission(IsAuthenticated):
+    def has_ipaddress_permission(self, request: Request, view: GenericAPIView, validated_serializer: Serializer):
+        user = User.from_request(request)
+        if (user.is_mreg_superuser_or_admin or user.is_mreg_network_admin):
+            return True
+        
+        if not hasattr(validated_serializer, "validated_data"):
+            return True
```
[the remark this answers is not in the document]
pederhan: It's guarding against being applied on classes that do not implement those fields.

24 views in total inherit from `HostPermissionsListCreateAPIView` and `HostPermissionsUpdateDestroy`. These two parent classes implement the new permission class `IsGrantedReservedAddressPermission`, but not all child views actually have an `ipaddress` field.

mreg/api/permissions.py
```diff
@@ -317,3 +358,48 @@ def has_destroy_permission(self, request, view, validated_serializer):
...
+
+        data = validated_serializer.validated_data   
+        if not data or not (ip := data.get("ipaddress")):
+            return True
```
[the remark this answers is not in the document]
pederhan: So yes, this specific check will pass when applied to an endpoint that does not implement a serializer with an `ipaddress` field. I don't think that's a bad thing, though. I don't think permission checks should deny based on assumptions. If the data does not contain an `ipaddress` field, the permission check has no more info to go on and should hand over the baton to the next permission check in line.

Consider this view:

https://github.com/unioslo/mreg/blob/65d64afe1464ca79ce3dece3cfcfab7bd4356e64/mreg/api/v1/views.py#L252-L254

We want users to have a matching permission triplet _and_ if, and only if, the data contains an ip address in an existing mreg network, the user needs network admin privileges to use a reserved address.

If we had used OR (`|`):

```py
class HostPermissionsUpdateDestroy(HostLogMixin, MregPermissionsUpdateDestroy):
    permission_classes = (IsGrantedNetGroupRegexPermission | IsGrantedReservedAddressPermission,)
```

Then any user would be able to use reserved addresses as long as they had netgroup regex permission, because the evaluation terminates on the first `True` it encounters. That's why we _need_ to use AND permissions. Hence why returning `False` on anything but an infringement of the actual rule we are checking seems wrong to me.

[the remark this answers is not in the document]
terjekv: Yeah, I see the point. The issue is that the check gets used on something it effectively isn't valid on, which means it has to return True, while in essence it should really return more of a "noop", ie, "don't use me for evaluating this clause". Returning True can be an issue if someone decides to use this as the only permissions check for something that will never match.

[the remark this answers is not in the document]
pederhan: Yes, I agree that it would be nice to have the ability to signal something else than `True/False` as well as determining the applicability of a given permission class on a view before/when applying it. However, in the absence of that, I think assuming `True` and returning `False` or raising an exception on specific infringements is easier to reason about than the opposite. I think it allows us to narrow down more effectively inside the actual methods.

We should also keep in mind that we implement this logic in custom methods that are checked _after_ `has_permission()` and `has_object_permission()`, and they all inherit from `IsAuthenticated`, so it's not like we forego all sorts of security by returning `True` in scenarios where we don't have enough info to go on.

I definitely see your point with regards to a deny-all approach, don't get me wrong. My gut feeling _does_ say that is the more secure approach, however, I feel like it's almost impossible to use and maintain permission classes that assume `False` by default.  Every valid permutation of user/query/data/view needs to be defined, as opposed to just denying the specific things we want to control for in that given permission class. Furthermore, A more lenient set of permission classes lets us compose them more easily with `|`, `!`, `,`, since otherwise we need to ensure each and every permission class is tailored to a given view.

PR #585, opened 2025-06-13
Log DRF errors, redact PermissionDenied details

pederhan:
## Issue

Whenever a user receives a 403 response, we are totally blind as to what went wrong. We have no logs of the error, and due to the use of predicates in permission classes, we are left in the dark as to _why_  the permission was denied. This makes it more or less impossible for us to determine at a glance why a user was forbidden from accessing a resource.

The default exception handler from `drf-standardized-errors` only logs server errors (5xx), and does not utilize structured logging. These two issues combined means we have no logs of any sort for client and validation errors.

## Solution

This PR extends the default `drf-standardized-errors` exception handler to log _all_ errors along with their stack traces and other useful context. Furthermore, it _guarantees_ that sensitive information is not propagated to 403 responses by always sending a generic 403. The error handler itself is still able to access the original 403 `PermissionDenied` exception, which lets us log it internally.

## Future work

The way I see it, in order to support more detailed logging of permission checks, the permission classes need be rewritten in  one of two ways:

1. Replace predicates (`return True/False`) with `raise PermissionDenied("info here")`
2. Split up/refactored into smaller, more specialized permission classes

Option 1 lets us log the exception info directly, but will require some work to ensure OR permissions `(permission_classes = (A | B,)` are handled correctly, since raised exceptions in `A` _could_ short-circuit internal DRF permission checks thereby never checking `B`. I have not verified whether this is the case.

Option 2 means we maintain the current DRF behavior by returning `False` to cause a `PermissionDenied` to be raised internally, but it will require extra work to ensure we log which permission class failed, as that is not clear from the stack trace by default.

Both approaches will require extending the internal DRF permission handling, but I think that is unavoidable at this point if we truly want more information about 403 errors.

### DIFF
--- a/mreg/exceptions.py
+++ b/mreg/exceptions.py
@@ -13,7 +13,7 @@ from rest_framework import exceptions
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from mreg.middleware.logging_http import get_request_user_agent, get_request_username
+from mreg.logs import get_request_method, get_request_user_agent, get_request_username
 
 mreg_logger = structlog.getLogger("mreg.http")
 
@@ -26,6 +26,8 @@ class MregExceptionHandler(ExceptionHandler):
 
     Furthermore, it redacts sensitive information from `PermissionDenied` exceptions
     to avoid leaking this information to users.
+
+    See: <https://drf-standardized-errors.readthedocs.io/en/latest/customization.html#handle-a-non-drf-exception>
     """
 
     def report_exception(self, exc: exceptions.APIException, response: Response) -> None:
@@ -40,7 +42,7 @@ class MregExceptionHandler(ExceptionHandler):
         # Extract useful request information
         request_username = get_request_username(request)
         request_user_agent = get_request_user_agent(request)
-        request_method = request.method if request and request.method else "UNKNOWN"
+        request_method = get_request_method(request)
         response_status = response.status_code
 
         # Get the stack trace from the original exception
--- a/mreg/logs.py
+++ b/mreg/logs.py
@@ -1,0 +1,49 @@
+"""Utility functions for logging."""
+
+from django.conf import settings
+from django.http import HttpRequest
+
+
+def get_request_header(request: HttpRequest, header_key: str, meta_key: str) -> str:
+    """Get the value of a header from the request, either via headers or META."""
+    if hasattr(request, "headers"):  # pragma: no cover
+        return request.headers.get(header_key)
+    return request.META.get(meta_key)
+
+
+def get_request_body(request: HttpRequest) -> str:
+    """Get the request body as a string, or '<Binary Data>' if it's binary.
+
+    We currently do not support multipart/form-data requests.
+    """
+    if request.POST:
+        return request.POST.dict()
+
+    try:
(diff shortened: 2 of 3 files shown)


mreg/log_processors.py
```diff
@@ -41,8 +41,8 @@ def filter_sensitive_data(_: Any, __: Any, event_dict: EventDict) -> EventDict:
     )
 
     if is_login_event:
-        content: str = event_dict["content"]
-        event: str = event_dict["event"]
+        content: str = event_dict.get("content", "")
+        event: str = event_dict.get("event", "")
```
pederhan, line 45: This caused an error in tests, since the new log record does not contain `content`.

mregsite/settings.py
```diff
@@ -46,7 +46,7 @@ def envvar(var: str, default: DefaultT) -> DefaultT:
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = ")e#67040xjxar=zl^y#@#b*zilv2dxtraj582$^(e6!wf++_n#"
 
-LOG_LEVEL = envvar("MREG_LOG_LEVEL", "CRITICAL").upper()
+LOG_LEVEL = envvar("MREG_LOG_LEVEL", "CRITICAL" if TESTING else "ERROR").upper()
```
pederhan, line 49: I welcome a better solution to not leak logs from 4xx errors to stderr during tests. Could maybe override the log handler inside the tests instead?

mreg/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    def report_exception(self, exc: exceptions.APIException, response: Response) -> None:
+        try:
+            drf_request: Request | None = self.context["request"]
+            request = drf_request._request  # pyright: ignore[reportOptionalMemberAccess, reportPrivateUsage]
+        except AttributeError:
+            # If the request is not available, we cannot log request-specific information
+            # NOTE: perform some sort of fallback logging here if needed
+            return super().report_exception(exc, response)
```
pederhan, line 38: I'm not sure how realistic this is. If we have a `Response` object, then _surely_ we have a `Request` as well? However, the type annotations for the context object don't guarantee the existence of a request, so here we are.

mreg/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+
+import traceback
+
+import structlog
+from drf_standardized_errors.handler import ExceptionHandler
+from rest_framework import exceptions
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from mreg.middleware.logging_http import get_request_user_agent, get_request_username
```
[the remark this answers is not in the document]
pederhan: I agree, they can live in a separate module (Name TBD. `log`, `logs`, `_logging` or `log_utils`?) that can be imported by both `exceptions` and `logging_http`.

I personally think "method" in the context of a request should be understood to mean HTTP method since these methods all take in a `HttpRequest` object.

pederhan:
This PR should probably be closed. We will get better auditing and introspection when we switch to policy engine-backed permissions, making this PR at best redundant, or more likely, obsolete.

PR #586, opened 2025-06-23
Add `/networks/?network__regex` filter

pederhan:
Fixes network regex lookup, which is used by the command `network find -network` in Mreg-cli.


terjekv (approved):
It works, so, eh, LGTM!

PR #589, opened 2025-08-05
Add support for /32 & /128 networks

pederhan:
Adds support for /32 (IPv4) and /128 (IPv6) networks, which was initially part of #584, but was later removed. This PR adds that functionality back.


terjekv (approved):
LGTM. Thanks!

PR #591, opened 2025-08-06
Reduce boilerplate in permissions tests

pederhan:
Adds helper function for creating a network and assigning permissions to it for one or more groups in permission tests.

### DIFF
--- a/mreg/api/v1/tests/test_permissions.py
+++ b/mreg/api/v1/tests/test_permissions.py
@@ -63,6 +63,20 @@ def get_mock_request(user: mock.Mock, mock_user_from_request: mock.Mock, path: s
     return request
 
 
+def create_network_and_permissions(
+    *groups: str,
+    network: str,
+    regex: str = r'.*\.example\.org$',
+    description: str = 'Test network'
+) -> Network:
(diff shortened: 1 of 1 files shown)


mreg/api/v1/tests/test_permissions.py
```diff
@@ -63,6 +63,20 @@ def get_mock_request(user: mock.Mock, mock_user_from_request: mock.Mock, path: s
     return request
 
 
+def create_network_and_permissions(
+    *groups: str,
```
terjekv, line 67: This syntax breaks my brain. I get it, but.. It's very magical and weird to read, but that might just be me?

pederhan: The alternative is always passing in a list of groups no matter what. I don't _love_ `*args`, but this is one of the few cases where it's acceptable in my mind.

terjekv: I just prefer the explicit type (I would be fine if we the syntax had been something like `*groups: list[str]` and have single arguments be coerced into a list, but seeing just `str` really hurts my head. Anyway, it's just me being silly.

PR #593, opened 2025-09-04
Copy .venv directory instead of using wheels in container image

pederhan:
This PR changes how the container image is built by relying completely on `uv sync` instead of using `pip` to build wheels. 

The changes in this PR are loosely based on https://docs.astral.sh/uv/guides/integration/docker/#non-editable-installs

By using uv the entire way through the build process, we should see some pretty significant speed-ups over the current `pip wheel` approach, but I have not measured this.

## New base image (`python:3.11-alpine`)

In order for symlinks to the system python used in the builder stage to transfer over to the final stage, the final stage has been modified to use `python:3.11-alpine` just like the builder. This could have also been solved with something like `uv python install 3.11` or similar, but why rely on uv to do this when there is already a base image with the Python version we want available?

## Size reduction

The total size of the final image has been reduced from ~275MB to ~200MB.


terjekv (approved):
LGTM, thanks!

pederhan:
CI fails because I merged master into this branch, which introduced PR image publishing via #601. Forks do not have access to the container registry, and thus they fail. Need to fix that, but it doesn't block this PR.

PR #616, opened 2026-02-01
Coverage updates post django6

terjekv:
Coverage work after the last set of major changes.


tox.ini
```diff
@@ -67,6 +70,7 @@ commands =
     python -m django --version
     coverage run --concurrency=multiprocessing manage.py test --parallel
     coverage combine
+    coverage run --append -m pytest -q mreg/api/tests
```
pederhan: Out of date after #654

tox.ini
```diff
@@ -54,6 +56,7 @@ commands = ruff check .
 
 [testenv:coverage]
 runner = uv-venv-lock-runner
+with_dev = true
```
pederhan: Out of date after #654

tox.ini
```diff
@@ -40,6 +41,7 @@ commands =
     !django52-!django60: python -m django version
     coverage run --concurrency=multiprocessing manage.py test --parallel
     coverage combine
+    coverage run --append -m pytest -q mreg/api/tests
```
pederhan: Out of date after #654

tox.ini
```diff
@@ -18,6 +18,7 @@ python =
 
 [testenv]
 runner = uv-venv-lock-runner
+with_dev = true
```
pederhan: Out of date after #654

pederhan (commented):
If I have one "criticism", it is that this PR performs HTTP calls differently than most of the existing tests: `self.client.<method>` along with an assertion on `response.status_code` is used over the "usual" `assert_<method>_and_<status>` methods.

Now,  I will say that I 100% prefer the way this PR goes about it, so I would rather see the rest of the test suite adapt to this PR's style in a follow-up PR.

mreg/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,385 @@
...
+    def test_delete_with_wrong_content_type(self):
+        """DELETE with a body and non-JSON content type is rejected."""
+        policy = NetworkPolicy.objects.create(name="test-policy")
+
+        response = self.client.generic(
+            "DELETE",
+            f"/api/v1/networkpolicies/{policy.pk}",
+            data=b"some body content",
+            content_type="text/html",
+        )
```
pederhan:
```suggestion
        response = self.client.delete(
            f"/api/v1/networkpolicies/{policy.pk}",
            data=b"some body content",
            content_type="text/html",
        )
```

This works in my local testing at least. Any reason `generic` was used here?

terjekv: Not really, I had some issues when fiddling with how to do these calls consistently due to problems with delete not supporting data at some point, and I think I gave halfway. I'll look into it.

terjekv:
> If I have one "criticism", it is that this PR performs HTTP calls differently than most of the existing tests: `self.client.<method>` along with an assertion on `response.status_code` is used over the "usual" `assert_<method>_and_<status>` methods.
> 
> Now, I will say that I 100% prefer the way this PR goes about it, so I would rather see the rest of the test suite adapt to this PR's style in a follow-up PR.

Yeah, I gave up debugging data passed through `assert_<method>_and_<status>`. If you look at that code there is so much boilerplate and more than slightly annoying to do inline debugging. I'm include to remove the lot of it as well.

PR #631, opened 2026-04-13
Fix local settings not configuring DB correctly

pederhan:
## Problem

The instructions in [README.md](https://github.com/unioslo/mreg/blob/bc8b6fa974b714d69c5198d38d4ab8cf6f77beb0/README.md#local-settings) and [local-docker-psql.MD](https://github.com/unioslo/mreg/blob/bc8b6fa974b714d69c5198d38d4ab8cf6f77beb0/local-docker-psql.MD) state that `DATABASES` should be redefined in order to use custom configurations. For most environments, this is overkill and actively harmful if not done correctly, as it overrides the default settings for activating [database connection pooling](https://docs.djangoproject.com/en/6.0/ref/databases/#connection-pool), which we define in settings.py:

https://github.com/unioslo/mreg/blob/bc8b6fa974b714d69c5198d38d4ab8cf6f77beb0/mregsite/settings.py#L240-L262

## Changes

This PR rewrites the documentation to use variable overrides in `local_settings.py` instead of redefining the entire database configuration. Furthermore, it moves the definition of `DATABASES` until after `local_settings.py` has been imported.

### Backwards compatibility

Backwards compatibility for older `local_settings.py` config files that define `DATABASES` has been added to prevent overriding those configurations with defaults:

https://github.com/unioslo/mreg/blob/e91b5c850d7cb6612cf773ca6287bdc7de6bb0d0/mregsite/settings.py#L451-L452

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ mreg supports configuration via environment variables with the `MREG_` prefix. T
 
 | Variable | Default | Description |
 | -------- | ------- | ----------- |
+| `MREG_DB_POOL_ENABLED` | `True` | Enable or disable database connection pooling |
 | `MREG_DB_POOL_MIN_SIZE` | `5` | Minimum idle connections in pool |
 | `MREG_DB_POOL_MAX_SIZE` | `25` | Maximum connections in pool |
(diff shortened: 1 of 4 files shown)


terjekv (approved):
LGTM

PR #635, opened 2026-07-02
Add `network_cidr` field to community serializer

pederhan:
This PR adds a new `network_cidr` field to serialized communities, which is the network CIDR of the network it is bound to. Improves API ergonomics.

## Rationale

In order to contact a community's endpoint, we are reliant on the network CIDR the community is bound to, i.e.

```
/networks/10.0.0.0/24/communities/1
          ^^^^^^^^^^^
```

But what we receive from the API for a Community object currently is something like this:

```jsonc
{
  "name": "mycommunity",
  "id": 1,
  "network": 123,
  // ...
}
```

Thus, consumers cannot construct the URL for the community based on just that object.

## New field: `network_cidr`

By adding the network CIDR to the serialized community, we can now construct the URL based exclusively on the community data without having to perform extra lookups or stitching together data from multiple objects:

```jsonc
{
  "name": "mycommunity",
  "id": 1,
  "network": 123,
  "network_cidr": "10.0.0.1/24",
  // ...
}
```

## Improves mreg-api ergonomics

Interacting with communities is awkward in mreg-api currently, because we need to manually fetch the network first, and pass that to every method that interacts with communities:

```py
# Fetch network, use network to fetch community
net = client.network.get("10.0.0.0/24")
com = client.network.community.get("mycom", net) 

client.network.community.update(com, net, description="My new description")
client.network.community.add_host(com, net, some_host_obj)
client.network.community.delete(com, net)
```

With this PR, the network CIDR is bound to the community object, and we can thus fetch the network as part of the community fetch operation (and , and we don't need pass around the network object:

```py
# No need to pass around network, can be fetched implicitly and discarded after fetching community
com = client.network.community.get("mycom", "10.0.0.0/24") 

client.network.community.update(com, description="My new description")
client.network.community.add_host(com, some_host_obj)
client.network.community.delete(com)
```

### DIFF
--- a/mreg/api/v1/serializers.py
+++ b/mreg/api/v1/serializers.py
@@ -35,9 +35,9 @@ class CommunitySerializer(serializers.ModelSerializer):
     # Members are all the hosts that have this community assigned.
     hosts = serializers.SerializerMethodField(read_only=True)
     global_name = serializers.SerializerMethodField(read_only=True)
-    network_address = serializers.SerializerMethodField(read_only=True)
+    network_cidr = serializers.SerializerMethodField(read_only=True)
 
-    def get_network_address(self, obj) -> str:
+    def get_network_cidr(self, obj) -> str:
         return str(obj.network.network)
 
     def get_hosts(self, obj):
@@ -81,7 +81,7 @@ class CommunitySerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Community
-        fields = ['id', 'name', 'description', 'network', 'network_address', 'hosts', 'global_name', 'created_at', 'updated_at']
+        fields = ['id', 'name', 'description', 'network', 'network_cidr', 'hosts', 'global_name', 'created_at', 'updated_at']
         read_only_fields = ['network'] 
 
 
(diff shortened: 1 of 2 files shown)


pederhan:
Need to change terminology to "CIDR", even though it's not used a lot throughout the codebase. We usually use the field name `network` for this kind of value, but that's already taken by the network ID in this case.

terjekv (approved):
For lenient clients this should be a fairly safe and sound extension. One thing is that `get_network_cidr()` always dereferences `obj.network` so unless we use `select_related("network")`, list/nested responses will probably add one network lookup per serialized community? How often do we serlalize out the community?

pederhan:
> How often do we serlalize out the community?

Only when fetching hosts, right? So this does affect endpoints that list hosts, because each host deserializes its community:

https://github.com/unioslo/mreg/blob/bbe8fa73d00fb7e8b3b7e06114c6f7519eb56a14/mreg/api/v1/views.py#L328-L335

Potential regression for `/hosts` endpoint? Particularly when listing _all_ hosts. Some regression tests would be _really_ nice. 

I'll look at this in late July/early August.

PR #636, opened 2026-07-04
Accept `primary_ns` & `nameservers` as strings in zone endpoints

pederhan:
This PR adds support for passing `primary_ns` as a string to zone endpoints by adding rudimentary type checking to `_get_request_nameservers()` to coerce string values to lists.

----

These endpoints pick the first item in the list passed to them, discarding any nameservers beyond the first one, so it makes sense to be able to pass this value as a string. Furthermore, this change makes the POST payload more consistent with the serialized representation of zones, which already serialize `primary_ns` as a string.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -152,8 +152,12 @@ jobs:
 
   publish:
     name: Publish
-    # Run for version tags, commits to master branch, and pull requests
-    if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/master' || github.event_name == 'pull_request'
+    # Fork pull requests cannot write organization packages with GITHUB_TOKEN.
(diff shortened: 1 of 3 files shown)


terjekv (approved):
LGTM, are there places we should document this, both mechanically and for humans?

pederhan:
> LGTM, are there places we should document this, both mechanically and for humans?

I am legitimately not sure. It took some digging to even understand where `primary_ns` was a list and where it was a string. I don't know if `primary_ns` is documented anywhere (it doesn't seem to be).

oyvindkolbu:
> > LGTM, are there places we should document this, both mechanically and for humans?
> 
> I am legitimately not sure. It took some digging to even understand where `primary_ns` was a list and where it was a string. I don't know if `primary_ns` is documented anywhere (it doesn't seem to be).

Oldest and the most arcane code in mreg. Could very well be cleaned up.

PR #637, opened 2026-07-04
Update Python dependencies

terjekv:
- Update uv-locked Python dependencies, including the open Dependabot bumps.
- Keep dependency updates split into one commit per changed dependency.
- Fix tox/uv Django fan-out so django52 and django60 environments install and assert the intended Django version.
- Confirmed django52 envs use Django 5.2.15 and django60 envs use Django 6.0.6.


pederhan (approved):
LGTM

PR #638, opened 2026-07-05
Fix not being able to disambiguate community hosts by IP when deleting

pederhan:


mreg/models/host.py
```diff
@@ -245,6 +245,18 @@ def _resolve_community_mapping(
...
+    def _resolve_ip(self, ip: Optional[Union['Ipaddress', str]] = None) -> Optional['Ipaddress']:
+        """
+        Helper method to resolve an IP address argument to an Ipaddress instance for the host.
+        """
+        if isinstance(ip, str):
+            try:
+                return Ipaddress.objects.get(host=self, ipaddress=ip)
+            except Ipaddress.DoesNotExist:
+                raise NotAcceptable("No IP address found on this host with the provided value.")
+        return ip
```
pederhan, line 258: Might as well ensure we provide the same interface for resolving IP args in both methods.

terjekv: That's just crazy talk

pederhan:
Do we need to do something here? 

https://github.com/unioslo/mreg/blob/3a51a184d500e752c1237074ea9f21aa33b6ae04/mreg/api/v1/serializers.py#L502-L530

My intuition says that we shouldn't touch that code. To me, it seems to make some unsound assumptions which it encodes through its own constraints checks...?

https://github.com/unioslo/mreg/blob/3a51a184d500e752c1237074ea9f21aa33b6ae04/mreg/api/v1/serializers.py#L493-L498

I also don't understand under which conditions this fires? When do we update a host with a `community` value that is not `None` but also evaluates to a falsey value (`0`, `""`, `[]`, etc.)?

mreg/api/v1/tests/test_network_policy.py
```diff
@@ -809,6 +809,34 @@ def test_delete_host_from_community_with_nonexistent_ipaddress_406(self):
...
+        )
+        ip1 = host.ipaddresses.first()  # type: ignore
+        ip2 = Ipaddress.objects.create(host=host, ipaddress="10.0.0.2")
+        self.addCleanup(ip2.delete)
+
+        # Ensure both IPs are actually on the same network
+        self.assertEqual(
+            Network.objects.get(network__net_contains=ip1.ipaddress),
+            Network.objects.get(network__net_contains=ip2.ipaddress),
+        )
```
pederhan, line 831: Maybe a little overzealous, but better safe than sorry!

mreg/models/host.py
```diff
@@ -296,17 +302,34 @@ def add_to_community(
...
+            if isinstance(community, Community):
+                mappings = HostCommunityMapping.objects.filter(host=self, community=community)
+            else:
+                mappings = HostCommunityMapping.objects.filter(host=self, community__name=community)
+            count = mappings.count()
+            if count == 0:
+                raise NotAcceptable("No community mapping exists for this host with the specified criteria.")
+            if count > 1:
+                raise NotAcceptable("Multiple IP addresses are mapped to this community; please specify one.")
+            mappings.delete()
```
terjekv: I thoughts about this. Should we avoid `mappings.count()` followed by `mappings.delete()`? I *think* Django keeps mapping lazy even after `count()` so the final `mappings.delete()` will run the query again. If someone has inserted anything that matches the filter, it'll be deleted by mistake? We should maybe do something ugly like this?

```python
mappings = list(mappings[:2])
if not mappings:
    raise NotAcceptable("No community mapping exists for this host with the specified criteria.")
if len(mappings) > 1:
    raise NotAcceptable("Multiple IP addresses are mapped to this community; please specify one.")

mappings[0].delete()
```

terjekv (approved):
Looks good enough to me.

PR #628, opened 2026-04-08
Fix OpenAPI schema generation

pederhan:
This PR fixes OpenAPI schema generation by switching from the built-in DRF OpenAPI schema generator to [drf-spectacular](https://drf-spectacular.readthedocs.io/en/latest/readme.html), as outlined in #627.

By using drf-spectacular, we get the following endpoints:

- `/docs`: Swagger UI for viewing the schema (default).
- `/docs/redoc`: ReDoc UI for viewing the schema (alternative).
- `/docs/schema`: The OpenAPI schema (YAML file) (`/docs` kept for backwards compatibility).

Since we can't assume MREG can access the internet to fetch static files for serving swagger/redoc UI pages, we include them manually via the `drf-spectacular[sidecar]` extra. Because both Swagger and ReDoc static files come from the same package, including endpoints for both costs nothing extra in terms of final container image size.

## Issues

Using drf-spectacular highlights numerous issues with our views:

<details>
  <summary><code>❯ uv run ./manage.py spectacular --color --file schema.yml</code></summary>

```
mreg/api/views.py:264: Error [HealthHeartbeat]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
mreg/api/views.py:274: Error [HealthLDAP]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
mreg/api/views.py:243: Error [MetaVersions]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
mreg/api/views.py:315: Error [MetricsView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
mreg/api/views.py:161: Error [UserInfo]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
[...]

PR #641, opened 2026-07-28
Fix stale pull request CI compatibility

terjekv:
## Summary

  - Preserve the pre-#628 field order for `NetworkSerializer` and `NetGroupRegexPermissionSerializer`.
  - Retain the explicit `network` and `range` fields required for OpenAPI generation.
  - Restore `mreg.api.views.LIBRARIES_TO_REPORT` as a compatibility alias.
  - Ensure mreg-cli tests the image built for the current pull request.
  - Add regression tests for both serializer orders and the compatibility alias.

  ## Root cause

  ### Serializer field ordering

  PR #628 explicitly declared two custom model fields so drf-spectacular could describe them:

  ```python
  NetworkSerializer.network = serializers.CharField()
  NetGroupRegexPermissionSerializer.range = serializers.CharField()
  ```

  Combined with `fields = "__all__"`, DRF moved those fields earlier in the serialized output.

  `NetworkSerializer` changed from:

  ```text
  id, excluded_ranges, policy, communities, created_at, updated_at, network, ...
  ```

  to:

  ```text
  id, network, excluded_ranges, policy, communities, created_at, updated_at, ...
  ```

  `NetGroupRegexPermissionSerializer` changed from:

  ```text
  id, created_at, updated_at, group, range, regex, labels
  ```

  to:

  ```text
  id, range, created_at, updated_at, group, regex, labels
  ```

  Although JSON object order is not semantically significant, mreg-cli’s recorded-output tests compare it, and stable serialization avoids unnecessary changes for existing consumers.

  Both serializers now use explicit field tuples in their previous order while retaining the explicit custom fields required for OpenAPI generation.

  ### Renamed constant

  PR #628 moved the reported-library constant from `mreg.api.views` to `mreg.api.serializers` and renamed it to `REPORTED_LIBRARY_VERSION_FIELDS`.

  In-flight pull requests created before #628, including #616, still import `LIBRARIES_TO_REPORT` from `mreg.api.views`. The previous name is retained as an alias:

  ```python
  LIBRARIES_TO_REPORT = REPORTED_LIBRARY_VERSION_FIELDS
  ```

  This keeps older branches compatible without duplicating the value.

  ### Incorrect image used by mreg-cli

  The container workflow tagged the newly built PR image locally as:

  ```text
  ghcr.io/unioslo/mreg:latest
  ```

  However, the mreg-cli integration script defaults to:

  ```text
  ghcr.io/unioslo/mreg:master
  ```

  The integration job could therefore test the published registry image instead of the image built for the current pull request.

  The workflow now tags the locally built image as `:master`, ensuring mreg-cli exercises the PR artifact. This local tag is not pushed to the registry.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -75,7 +75,7 @@ jobs:
     - name: Tag container image
       # There's a docker-compose.yml file in the mreg-cli repo that wants the image from ghcr.io,
       # but we want to use the newly built custom image
-      run: docker tag mreg ghcr.io/unioslo/mreg:master
+      run: docker tag mreg ghcr.io/unioslo/mreg:latest
     - name: Checkout
       uses: actions/checkout@v4
     - name: Download mreg-cli
@@ -89,6 +89,8 @@ jobs:
         git -c advice.detachedHead=false checkout $C
         cp --no-clobber /tmp/ci/* ci/
     - name: Run the tests
+      env:
+        MREG_IMAGE: ghcr.io/unioslo/mreg:latest
       run: mreg-cli/ci/run_testsuite_and_record_V2.sh
     - name: Comment on the commit that the tests worked
       run: |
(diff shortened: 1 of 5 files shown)


pederhan (approved):
LGTM. We must look into more robust and flexible diffing in mreg-cli ([DeepDiff](https://zepworks.com/deepdiff/current/index.html)), so we avoid this churn.

PR #644, opened 2026-07-29
CI: use the local MREG image for mreg-cli tests

terjekv:
## Summary

- give the workflow-built MREG image a clear local CI tag
- configure the mreg-cli compatibility test to prohibit registry pulls
- keep the default pull-request merge checkout so CI tests the would-be merge result

## Why

Docker Compose always pulls images tagged `latest`, so the compatibility job could replace the workflow-built image with a registry image. The local `mreg:ci` tag avoids the special `latest` behavior, while `MREG_IMAGE_PULL_POLICY=never` makes the local-only requirement explicit through unioslo/mreg-cli#467.

## Validation

- `actionlint -shellcheck "" .github/workflows/container-image.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -65,6 +65,10 @@ jobs:
     name: Test with mreg-cli
     needs: build
     runs-on: ubuntu-latest
+    env:
+      # This image exists only in the runner's local Docker image store.
+      MREG_IMAGE: mreg:ci
+      MREG_IMAGE_PULL_POLICY: never
     steps:
     - name: Download artifact
       uses: actions/download-artifact@v8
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image.yml
```diff
@@ -65,17 +65,19 @@ jobs:
     name: Test with mreg-cli
     needs: build
     runs-on: ubuntu-latest
+    env:
+      # This image exists only in the runner's local Docker image store.
+      MREG_IMAGE: mreg:ci
+      MREG_IMAGE_PULL_POLICY: never
```
pederhan, line 72: Should probably mention that these env vars control the behavior of [run_testsuite_and_record_V2.sh](https://github.com/unioslo/mreg-cli/blob/master/ci/run_testsuite_and_record_V2.sh) so it doesn't look so magical?

```suggestion
    env:
      # These vars are sourced by mreg-cli's `run_testsuite_and_record_V2.sh`
      # Ensures we use local image (with unique CI tag), not image pulled from registry.
      MREG_IMAGE: mreg:ci
      MREG_IMAGE_PULL_POLICY: never
```

pederhan (approved):
LGTM

